PR #11, opened 2023-09-01
refactor: dispatch collection metadata jobs on demand when collection is viewed

crnkovic:
## Summary

https://app.clickup.com/t/865cyr0k6

## Checklist

-   [ ] Ready to be merged

### DIFF
--- a/tests/App/Jobs/FetchCollectionVolumeTest.php
+++ b/tests/App/Jobs/FetchCollectionVolumeTest.php
@@ -40,3 +40,9 @@ it('has a retry limit', function () {
 
     expect((new FetchCollectionVolume($collection))->retryUntil())->toBeInstanceOf(DateTime::class);
 });
+
+it('has a unique ID', function () {
+    $collection = Collection::factory()->create();
+
+    expect((new FetchCollectionVolume($collection))->uniqueId())->toBeString();
+});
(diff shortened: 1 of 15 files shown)


tests/App/Jobs/FetchCollectionVolumeTest.php
```diff
@@ -40,3 +40,9 @@
 
     expect((new FetchCollectionVolume($collection))->retryUntil())->toBeInstanceOf(DateTime::class);
 });
+
+it('has a unique ID', function () {
+    $collection = Collection::factory()->create();
+
+    expect((new FetchCollectionVolume($collection))->uniqueId())->toBeString();
+});
```
shahin-hq, line 48: Don't you think this test need to be refined as we already have return type of the `uniqueId` method:

```PHP
public function uniqueId(): string
```

crnkovic: don't think it brings any value. this test is purely to satisfy 100% coverage. should changing the implementation of a `uniqueId` to something else make the tests fail, even if the updated value ensures unique jobs? the point is to test what it does (uniqueness), not the actual implementation that might change

shahin-hq: 1. the point is to test what it does (uniqueness), not the actual implementation that might change
2.  increasing the minutes from 10 to 15 is not crucial enough that should fail the tests

I do agree both of these statements, however, changing `uniqueId` implementation to something `hello-world` should make the test fail, I think. Even making it unique also should make the test fail just to warn  dev that you might mess up stuff, be careful and ensure that test is fixed. What do you think?

shahin-hq (approved):

oXtxNt9U (approved):

oXtxNt9U:
@crnkovic merge conflicts

PR #17, opened 2023-09-04
fix: ignore watch pattern on pnpm build

ItsANameToo:
## Summary

https://app.clickup.com/t/861na8a4k

It now checks for `NODE_ENV`, which will be `production` when `pnpm build` is called (it's `development` for `pnpm dev`)

This means that any build call will ignore the watcher, which should be ok as we run `pnpm dev` during development and that's when files may change, so it shouldn't affect any regular building scenarios. Open to insights from others though

## Checklist

-   [ ] Ready to be merged

### DIFF
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -34,10 +34,12 @@ export default defineConfig({
             input: "resources/js/app.tsx",
             refresh: true,
         }),
-        watch({
-            pattern: "app/{Data,Enums}/**/*.php",
-            command: "npm run generate:typescript",
-        }),
+        process.env.NODE_ENV !== "production"
+            ? watch({
+                  pattern: "app/{Data,Enums}/**/*.php",
+                  command: "npm run generate:typescript",
+              })
+            : null,
         react(),
         svgr(),
         markdown(),


vite.config.ts
```diff
@@ -34,10 +34,12 @@ export default defineConfig({
...
-        watch({
-            pattern: "app/{Data,Enums}/**/*.php",
-            command: "npm run generate:typescript",
-        }),
+        process.env.NODE_ENV !== "production"
+            ? watch({
+                  pattern: "app/{Data,Enums}/**/*.php",
+                  command: "npm run generate:typescript",
+              })
+            : null,
```
oXtxNt9U, line 42: Not sure if a `null` plugin will cause any issues, but we could also do it like this:

```suggestion
        ...(process.env.NODE_ENV !== "production"
            ? [watch({
                pattern: "app/{Data,Enums}/**/*.php",
                command: "npm run generate:typescript",
            })]
            : []
        ),
```

oXtxNt9U (approved):
Looks fine to me, just left one comment but will already approve

PR #18, opened 2023-09-04
chore: update PHP dependencies

ItsANameToo:
Automated changes by [create-pull-request](https://github.com/peter-evans/create-pull-request) GitHub action


shahin-hq:
@ItsANameToo should we ignore offended rector?

ItsANameToo:
@shahin-hq we can pin rector to 0.17.6 again as it was before ( `"rector/rector": "0.17.6",` ) as they are changing a lot in their recent updates. Then that can be updated separately from this PR to also refactor anything that's needed without it causing a block on this PR to update the other deps

samharperpittam (approved):

PR #19, opened 2023-09-04
fix: remove duplicate collections list items when changing collection visibility status

goga-m:
## Summary
Closes https://app.clickup.com/t/8678j57ra

## Checklist

-   [x] Ready to be merged


crnkovic:
Hm interesting. How does this solve a problem when IDs are already unique in the database? 🤔

ItsANameToo:
> @crnkovic it will render as e.g `220-0`, `220-1` ..., so I think the keys would be unique even if ids are the same 🤔

but `id` is unique, that's database id so you don't get `220-0`, `220-1` to begin with? 🤔

crnkovic:
as if the database returns duplicated NFTs, i can see it throwing an error in that case

ItsANameToo:
Although it does not show the error in devtools anymore on this branch, is does not update the table properly either 🤔 I hid one and it still shows up in the table when it refreshes to hide it. Clicking the details button on it causes it to show "unhide" while it's in the regular list view. Toggling `show hidden` is also broken at that point, as it will show the same table until you refresh the page
image

ItsANameToo:
to reproduce:
- use a wallet with many collections
- scroll down to load new collections
- hide one after it loaded the images in that list of new collections
- no errors, but collection will remain and toggle no longer works as expected

shahin-hq (approved):

ItsANameToo (approved):

ItsANameToo:
conflict @goga-m

PR #48, opened 2023-09-07
refactor: switch off  `owned` toggle if user owns zero nfts

goga-m:
## Summary

Closes https://app.clickup.com/t/862kcdy7a

## Checklist

-   [ ] Ready to be merged


samharperpittam (changes requested):
When I view a Collection page with a wallet that owns 0 NFTs for that Collection, the Collection page doesn't render NFT's until I toggle "Show My Collection" twice. 

See video

https://github.com/ArdentHQ/dashbrd/assets/44671015/7d937a15-3847-46e5-bd42-75eaa7160a7e

samharperpittam (approved):

resources/js/Pages/Collections/View.tsx
```diff
@@ -90,7 +90,7 @@ const CollectionsView = ({
 
         return {
             traits: hasSelectedTraits ? selectedTraits : undefined,
-            owned: showOnlyOwned ? undefined : false,
+            owned: showOnlyOwned,
```
alfonsobries, line 93: With this adjustment it always add the `owned` param in the url, see video

I think that we should keep the default value as `undefined`, is not a breaking change per se and is not causing any issue but more like a personal preference since in this way we can prevent noise on the url, 



https://github.com/ArdentHQ/dashbrd/assets/17262776/f7d43b13-5ed5-49e6-afba-9bb94af40019

alfonsobries (changes requested):
If you open a collection were you actually have nfts it shows your collection as expected, but you cannot longer hide your collections, on the screenshot show my collection is disabled but i still see my collection

image

ItsANameToo:
> If you open a collection were you actually have nfts it shows your collection as expected, but you cannot longer hide your collections, on the screenshot show my collection is disabled but i still see my collection
> 
> image

this seems unrelated to the PR as the same happens on develop 🤔

ItsANameToo (approved):

resources/js/Pages/Collections/View.tsx
ItsANameToo: this is needed after all to avoid the toggle from not hiding all NFTs when you own 0

PR #56, opened 2023-09-08
refactor: store banner for collections when fetching wallet NFTs

patricio0312rev:
# [[collections] look into banner issue](https://app.clickup.com/t/861naud7x)

- The collection's banner wasn't being displayed when we initially fetched NFTs when connecting a new wallet.
- The query we use for that: `getNfts`, returned the banner as a property inside `metadata.openSea.banner` as [mentioned in their API reference](https://docs.alchemy.com/reference/getnfts).
- To solve this problem, we're simply adding the `bannerImageUrl` as an optional property in the `Web3NftData` object.

image

## Steps to reproduce
1. Run the app in `test_e2e` mode.
2. Run `php artisan horizon`.
3. Make sure your testing wallet has multiple NFTs from different collections.
4. Go to `/collections` page, select any Collection and see the magic ✨ 

## Checklist

-   [ ] Ready to be merged


app/Data/Web3/Web3NftData.php
```diff
@@ -28,6 +28,7 @@ public function __construct(
         public ?string $collectionImage,
         public ?string $collectionWebsite,
         public ?string $collectionDescription,
+        public ?string $bannerImageUrl,
```
ItsANameToo, line 31: would rename it to `collectionBannerImageUrl` so it's in line with the other collection items here to avoid confusion

ItsANameToo (changes requested):

ItsANameToo (approved):

PR #65, opened 2023-09-08
refactor: ignore banners on demand if banner already exists

patricio0312rev:
# [[general] ignore banners on demand if banner already exists](https://app.clickup.com/t/861nax830)

## Summary

- Frequency for `FetchCollectionBanner` job has been changed from daily to weekly.
- Collections that already have a banner won't run that job on demand.

## Checklist

-   [ ] Ready to be merged


app/Console/Kernel.php
```diff
@@ -132,13 +132,12 @@ private function scheduleJobsForCollections(Schedule $schedule): void
...
 
         // Fetch banners for collections that don't have one yet
-        // (More often according to config)
         $schedule
             ->command(FetchCollectionBanner::class, [
                 '--missing-only',
             ])
             ->withoutOverlapping()
-            ->dailyAt('5:30');
+            ->weekly();
```
ItsANameToo, line 140: I'd keep this at daily as it's only for missing banners, which makes sense to try and fetch more often as new collections may add one shortly after we fetch them

app/Http/Controllers/CollectionController.php
```diff
@@ -144,6 +145,10 @@ public function view(Request $request, Collection $collection): Response
         $reportAvailableIn = RateLimiterHelpers::collectionReportAvailableInHumanReadable($request, $collection);
 
         if (! $collection->recentlyViewed()) {
+            if ($collection->banner() === null) {
```
ItsANameToo, line 148: here I'd add a check of a week before refetching. Otherwise this change will cause the banners to not update at all if they existed prior, which means that any changes to banners would not be picked up

ItsANameToo (changes requested):

app/Http/Controllers/CollectionController.php
ItsANameToo: yeah would need something specific for the banner to avoid it being skipped when other aspects of the model get updated

app/Models/Collection.php
```diff
@@ -111,6 +111,11 @@ public function banner(): ?string
         return $this->extra_attributes->get('banner');
     }
 
+    public function bannerUpdatedAt(): ?string
+    {
+        return $this->extra_attributes->get('banner_updated_at');
```
shahin-hq, line 116: I'd parse `banner_updated_at` right here, so we won't have to deal with parsing in multiple places.

oXtxNt9U (approved):

ItsANameToo (approved):

PR #68, opened 2023-09-11
refactor: fetch activity for the entire collection as opposed to each nft

crnkovic:
## Summary

https://app.clickup.com/t/861nap53e

In this PR, we move away from fetching activity for each individual NFT, to fetching activity for entire collections. This greatly helps with performance as we suspect there are many, many NFTs that have no recent activity. Periodically running the job to fetch their activity is a waste of both the server resources and API quota as many of them will report no new activity.

By fetching the activity on the collection level, we can avoid many unnecessary requests and jobs being pushed to the queue. For example, if a collection has hundreds of NFTs, but none of those NFTs has any activity, this would be 1 job/request to ensure no new activity has happened for all those hundreds of NFTs.

We link NFTs with their respective activity by the collection ID and their token ID instead of the internal NFT ID as it is now.

As some collections have millions of NFTs and millions of activity items, we can prefetch the entire activity log for these collections and manually insert them into the database, avoiding unnecessary jobs/requests taking hours to index everything on the production server. Once we have all the activity up to a certain point in time (in the near past), we can schedule the job to sync the latest activity. I don't assume there will be a lot of recent activity that's too hard for us to handle, but we should fine-tune this as we have more actual data on how active collections are. For more details on data migration, look for comments on the ClickUp card.

### Code changes

- Instead of using `nft_id` in the `nft_activity` table, we now use `collection_id` and `token_number`. This allows us to prefetch the activity for the NFTs that do not exist in our database. Then we use those values as keys for `activities` relationship on the NFT model ([diff](https://github.com/ArdentHQ/dashbrd/pull/68/files#diff-2940177b725bbc7f9c7a88c08246ce65f920365d22a9c8a88d04383ccf30ccfcR93)).
- When the collection is created in the database, we dispatch a `FetchCollectionActivity` job. This ensures we index the activity for newly created collections.
- This PR adds the `is_fetching_activity` boolean column to collections. As fetching entire activity log can take hours, we want to ignore those collections that are still getting indexed from the scheduler. In essence, let it index in peace.
- Rest of the logic is the same. We use the same Mnemonic endpoint (omitting the `tokenId` parameter) to fetch the activity. We `UPSERT` them, and if we get the limit (500 items) we dispatch another job in case there are more items to retrieve. Once we get less than 500 items, we can assume there is no more activity items to fetch.
- By using timestamp as a cursor, `FetchCollectionActivity` job is restitant to errors. If the job is interrupted (for example failed job), the next time it runs it will just take the latest timestamp activity and start from that point.

### Retrieving the latest activity only for a specific NFT

Since the job to fetch the collection activity uses the timestamp of the newest activity item as a cursor, dispatching a job to fetch the activity for a specific NFT would break both this logic (timestamp cursor would move) and sorting ("Recently Received") would be out of date as only some NFTs have their activity updated. So I opted not to include that in this PR at all. The only way to update the activity for a specific NFT is to update the activity for the entire collection.

### Disabling activity indexing

There are times when we just want to disable indexing activities, as this takes a long time. This can be for demo, as we have a good set of data (10M) to test everything, or during the importing of production data. In that case, you can set `ACTIVITIES_ENABLED=false` environment variable.

### Ignoring specific collections from having their activity indexed

If you want to ignore only some collections from having their activity indexed, you can update the `activity_blacklist` array to include the address of a collection in the  `config/dashbrd.php` config file.

## Setup

Our drive contains 2 files, `demo.sql.gz` and `prod.sql.gz`. Those contain activities data compressed into the archive. Workflow for this PR:

1. Set `ACTIVITIES_ENABLED=false` environment variable so activities indexer doesn't trigger
2. Truncate the `nft_activity` table. Make sure to restart the identity so the primary key sequence (ID) restarts. TablePlus has this option when truncating, which is how I did it.
3. Deploy the PR and run migrations. You can repeat the truncation after this step just to be sure it's empty.
4. Download gzips locally from drive and scp them onto the server into the `/tmp` directory: `scp /path/to/demo.sql.gz forge@ip:/tmp`
5. SSH in the server, log in as `postgres` user and unzip the archive: `sudo su postgres && cd /tmp && gzip -k -d demo.sql.gz`
6. Import the dump into the database (for prod this took around 5mins for me): `psql -U postgres -d database_name < demo.sql` (replace `database_name` with `forge` or whatever the database name is)
7. Ensure it imported everything after it finishes
8. Delete `demo.sql` and `demo.sql.gz` files
9. Enable activity indexing: `ACTIVITIES_ENABLED=true` and restart Horizon with `php artisan horizon:terminate`. **Ideally, we'd leave this disabled for demo as there's no need to keep indexing activities and SQL dump doesn't contain all activities.**

If importing the dump into the demo fails, let me know. This will only happen if some of the records from the `collections` table have been removed, since we now match the `collection_id` from the dump to the `id` on the `collections` table. Our drive contains the CSV containing collections for both demo and prod as well.

## Testing

For locally testing the PR, you don't need anything special. Set the `ACTIVITIES_ENABLED=true` and run `php artisan collections:fetch-activity --collection-id=1` or whatever ID you want. Make sure to pass the ID or it will run for every collection and will take a while.

## Checklist

-   [ ] Ready to be merged


app/Jobs/FetchCollectionActivity.php
```diff
@@ -0,0 +1,125 @@
...
+
+        if ($this->collection->is_fetching_activity && ! $this->forced) {
+            return;
+        }
+
+        $this->collection->update([
+            'is_fetching_activity' => true,
+        ]);
+
+        $limit = 500;
```
alfonsobries, line 54: I would move this value to constant

app/Jobs/FetchCollectionActivity.php
```diff
@@ -0,0 +1,125 @@
...
+                'total_usd' => $activity->totalUsd,
+                'extra_attributes' => json_encode($activity->extraAttributes),
+            ])->toArray();
+
+        DB::transaction(function () use ($formattedActivities, $activities, $limit) {
+            NftActivity::upsert($formattedActivities, uniqueBy: ['tx_hash', 'collection_id', 'token_number', 'type']);
+
+            // If we get the limit it may be that there are more activities to fetch...
+            if ($limit === count($activities)) {
+                self::dispatch($this->collection, forced: true)->afterCommit();
```
alfonsobries, line 95: calling a event inside a db transaction kinda bothers me, doesnt need to create any error but i think does not belongs here, what do you think?

app/Jobs/FetchCollectionActivity.php
```diff
@@ -0,0 +1,125 @@
...
+                    'activity_updated_at' => now(),
+                ]);
+            }
+        }, attempts: 5);
+    }
+
+    private function latestActivityTimestamp(): ?Carbon
+    {
+        return $this->collection
+                    ->activities()
```
alfonsobries, line 108:
```suggestion
                    ->activities()
                    ->select('timestamp')
```

in theory that should make the query faster

database/migrations/2023_09_11_071352_add_activity_updated_at_column_to_collections_table.php
```diff
@@ -0,0 +1,21 @@
...
+
+return new class extends Migration
+{
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        Schema::table('collections', function (Blueprint $table) {
+            $table->boolean('is_fetching_activity')->default(false)->after('minted_at');
```
oXtxNt9U, line 17: I guess you could add an index on this column

database/migrations/2023_09_07_135157_add_token_number_column_to_nft_activity_table.php
```diff
@@ -0,0 +1,27 @@
...
+    public function up(): void
+    {
+        Schema::table('nft_activity', function (Blueprint $table) {
+            $table->addColumn('numeric', 'token_number', ['numeric_type' => 'numeric'])->nullable();
+            $table->foreignIdFor(Collection::class)->nullable()->constrained()->cascadeOnDelete();
+            $table->unsignedBigInteger('nft_id')->nullable()->change();
+
+            // TODO: remove `nft_id`
+
+            $table->unique(['tx_hash', 'collection_id', 'token_number', 'type']);
```
oXtxNt9U, line 24: Here we also need to include `log_index`. See [this](https://app.clickup.com/t/85ztyhwq4) task.

app/Jobs/RefreshNftMetadata.php
```diff
@@ -39,7 +38,7 @@ public function __construct(
      */
     public function handle(AlchemyWeb3DataProvider $provider): void
     {
```
oXtxNt9U, line 40: Something we can consider is to also dispatch a `FetchCollectionActivity` job here, so it's possible to manually refresh specific collections without having to wait for the backend.

app/Jobs/FetchCollectionActivity.php
crnkovic: interestingly this produces the same SQL query :noice:

```
> Collection::find(3)->activities()->latest('timestamp')->select('timestamp')->value('timestamp')
= select "timestamp" from "nft_activity" where "nft_activity"."collection_id" = ? and "nft_activity"."collection_id" is not null order by "timestamp" desc limit 1  

> Collection::find(3)->activities()->latest('timestamp')->value('timestamp')
= select "timestamp" from "nft_activity" where "nft_activity"."collection_id" = ? and "nft_activity"."collection_id" is not null order by "timestamp" desc limit 1  

```

app/Jobs/FetchCollectionActivity.php
```diff
@@ -0,0 +1,125 @@
...
+    {
+        $this->collection->update([
+            'is_fetching_activity' => false,
+            'activity_updated_at' => now(),
+        ]);
+    }
+
+    public function retryUntil(): DateTime
+    {
+        return now()->addMinutes(10); // This is retry PER JOB (i.e. per request)...
```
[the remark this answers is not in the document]
crnkovic: This basically dispatches another job. The fact it's the same type of job is just a "coincidence." Interesting, will test it. I reckon the `retryUntil` should be per job, not per job-type.

app/Jobs/FetchCollectionActivity.php
crnkovic: I just did that to avoid repetition as we now have simple if-else. It still has `afterCommit` so I don't think it would be a problem. The alternative approach is:

```php
DB::transaction(function () use ($formattedActivities, $activities, $limit) {
    NftActivity::upsert($formattedActivities, uniqueBy: ['tx_hash', 'collection_id', 'token_number', 'type']);

    // If we get the limit it may be that there are more activities to fetch...
    if ($limit !== count($activities)) {
        $this->collection->update([
            'is_fetching_activity' => false,
            'activity_updated_at' => now(),
        ]);
    }
}, attempts: 5);

if ($limit === count($activities)) { // repeating
    self::dispatch($this->collection, forced: true);
}
```

It's still basically the same code, both ways are just fine in my book, what do you think?

app/Data/Web3/CollectionActivity.php
```diff
@@ -19,11 +19,20 @@ public function __construct(
...
         public array $extraAttributes,
     ) {
     }
+
+    public function key(): string
+    {
+        return implode(':', [
+            $this->txHash,
+            $this->tokenId,
+            $this->type->value,
```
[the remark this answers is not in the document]
crnkovic: Since this PR changes a database structure a bit, I pushed a commit to add `log_index` column as well and have deduplicated also by that attribute 🤷

app/Console/Kernel.php
```diff
@@ -96,6 +97,11 @@ protected function schedule(Schedule $schedule): void
 
     private function scheduleJobsForCollectionsOrGalleries(Schedule $schedule): void
     {
+        $schedule
+            ->command(FetchCollectionActivity::class)
+            ->withoutOverlapping()
+            ->dailyAt('10:00');
```
ItsANameToo, line 103: I don't think we need to do this daily for all collections. I'd start with weekly and we should fetch it more often for collections where we notice people more often view them or if they are more popular in general (most owners in db, some other popularity metric, etc) . Out of scope for this PR, so I'd just start with weekly updates for now

app/Http/Controllers/RefreshedNftController.php
```diff
@@ -21,8 +20,6 @@ public function __invoke(Collection $collection, Nft $nft): JsonResponse
         // It's supposed to be completely opaque to the user what the "refresh" is doing.
         RefreshNftMetadata::dispatch($collection, $nft)->onQueue(Queues::NFTS);
 
-        FetchNftActivity::dispatch($nft);
```
ItsANameToo, line 24: I'd still fetch it for the entire collection so it can be updated on a different interval if needed. This is mainly to allow to change activity to be a weekly scheduled job; then we run the risk of some activities being behind and we should allow the users to update them once a day (1x per day per collection max, irregardless of user requesting it)

crnkovic:
@ItsANameToo there are 2 collections that have a low supply, but a lot of activity. i have deployed dashbrd on a random server and has been running to index those, and all other activities. i basically have staging activities ready and we can test on staging, for prod we can either ditch activities for those 2 collections or wait for them to index all :sweating:

ItsANameToo:
which collections are they

oXtxNt9U (approved):
Just a small suggestion (can ignore if you think it's unnecessary).

Looks good to me otherwise. Tested it locally, activities are fetched and show up.

ItsANameToo:
@crnkovic can you rebase this on `rc-0.8.0` so we can get it released separately from the stuff on `develop` please

ItsANameToo:
this seems to include changes from develop still @crnkovic

ItsANameToo:
superseded by #217 where it's rebased on the rc branch

PR #79, opened 2023-09-12
fix: avoid collection's last indexed token number becoming null

alfonsobries:
## Summary

Closes https://app.clickup.com/t/861nbc3f8

## Checklist

-   [ ] Ready to be merged

### DIFF
--- a/app/Support/Web3NftHandler.php
+++ b/app/Support/Web3NftHandler.php
@@ -51,7 +51,7 @@ class Web3NftHandler
         $collectionsData = $nftsGroupedByCollectionAddress->flatMap(function (Web3NftData $nftData) use ($now, $nftsInCollection) {
             $token = $nftData->token();
 
-            $data = [
+            return [
                 $nftData->tokenAddress,
                 $nftData->networkId,
                 trim($nftData->collectionName),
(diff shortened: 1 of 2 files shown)


app/Support/Web3NftHandler.php
```diff
@@ -99,8 +111,8 @@ public function store(Collection $nfts, bool $dispatchJobs = false): void
             floor_price_retrieved_at = excluded.floor_price_retrieved_at,
             extra_attributes = excluded.extra_attributes,
             minted_block = excluded.minted_block,
-            minted_at = excluded.minted_at,
-            last_indexed_token_number = excluded.last_indexed_token_number
+            {$lastIndexedTokenConflict}
```
oXtxNt9U, line 114: I think we can simplify this entire diff to just

```sql
last_indexed_token_number = coalesce(excluded.last_indexed_token_number, collections.last_indexed_token_number)
```

Meaning that if we get a non-null timestamp we use it, otherwise we use the existing timestamp in the collections table which has the same effect as the above conditionals.

oXtxNt9U (changes requested):

app/Support/Web3NftHandler.php
ItsANameToo: applied it, test seems to like it as well and makes sense

ItsANameToo (approved):

oXtxNt9U (approved):

PR #81, opened 2023-09-12
feat: add network filter to collections overview

patricio0312rev:
# [[collections] Add filter option per network](https://app.clickup.com/t/861nam9eh)

## Summary

- Collections can now be filtered by chain.

## How is it done?

- In the `view` function from `CollectionController`, a new filter for the `chain` query param has been added. 
- Here, we just use that query param to create an array of `chain_ids` and filter our collections in case we have 1 or more ids in the array.

## Changes

- New `NetworkWithCollectionsData` data constructor has been created to help us get the collections count for each network and get the correct typing.
- New translations have been added for the new component inside `pages.php`
- New `CollectionsNetworksFilter` component to contain all the filters.

> **_NOTE:_**  If you set `TESTNET_ENABLED` to `true` in your env vars, you will see the test networks in your filters.

https://github.com/ArdentHQ/dashbrd/assets/55117912/0badcf84-2ba9-449a-8416-44247dacbe69

## Checklist

-   [ ] Ready to be merged


samharperpittam (changes requested):
Quick change on this, when navigating to `/collections`, by default all collections are displayed across both eth and poly, but the filter shows both eth and poly deselected. 

Can we change this so the network options are selected by default. 
image

app/Data/Network/NetworkWithCollectionsData.php
```diff
@@ -0,0 +1,53 @@
...
+        public int $collectionsCount,
+    ) {
+    }
+
+    /**
+     * @return Collection<int, NetworkWithCollectionsData>
+     */
+    public static function fromModel(User $user, bool $showHidden)
+    {
+        $networks = Network::scopeOnlyActive()->get();
```
crnkovic, line 34: `Network::onlyActive()`

app/Data/Network/NetworkWithCollectionsData.php
```diff
@@ -0,0 +1,53 @@
...
+                $network->name,
+                $network->is_mainnet,
+                $network->chain_id,
+                $network->public_rpc_provider,
+                $network->explorer_url,
+                $collectionsPerNetwork[$network->id] ?? 0
+            );
+        });
+
+        return $networksWithCollectionsCount;
```
crnkovic, line 51: just return inline, no need for a new variable.

speaking of that, why create a map for networkID - counts, when you can just inline everything:

```php
return $networks->map(fn ($network) => new NetworkWithCollectionsData(
    $network->id,
    $network->name,
    $network->is_mainnet,
    $network->chain_id,
    $network->public_rpc_provider,
    $network->explorer_url,
    $user->collections()->where('network_id', $network->id)->when($showHidden, fn ($q) => $q->whereIn('collections.id', $user->hiddenCollections->modelKeys()))->count()
);)
```

or am i missing something?

crnkovic (changes requested):
The card specifies to disable the checkbox if there are no collections for the network, but I can still toggle it:

Screenshot 2023-09-14 at 14 51 08

Also, if you hide a collection, it will still display the total number of collections per network, including hidden ones. It should probably take into account only collections based on shown/hidden state (hidden = show only hidden count per network and vice versa). cc @samharperpittam.

Screenshot 2023-09-14 at 14 56 46

EDIT: Ah, I have noticed that the shown/hidden bug only happens if you re-toggle "showHidden". If I refresh the page, the counts are okay.

samharperpittam:
> Also, if you hide a collection, it will still display the total number of collections per network, including hidden ones. It should probably take into account only collections based on shown/hidden state (hidden = show only hidden count per network and vice versa). cc @samharperpittam.

Yeah I agree with this

crnkovic:
Something is screwed up with counts when showing/hiding 😂 

https://github.com/ArdentHQ/dashbrd/assets/6536260/d47cbe79-5025-4c78-ab27-fceab789d3b5

crnkovic (approved):

app/Data/Network/NetworkWithCollectionsData.php
ItsANameToo: he provided a solution already unless I'm missing something @patricio0312rev 🤔

resources/js/Pages/Collections/Components/CollectionsFilter/CollectionsNetworksFilter.tsx
```diff
@@ -0,0 +1,83 @@
...
+    count,
+    value,
+    isSelected,
+}: {
+    handleClick: (value: number) => void;
+    count: number;
+    value: App.Data.Network.NetworkWithCollectionsData["chainId"];
+    isSelected: boolean;
+}): JSX.Element => (
+    <label className="group flex cursor-pointer items-center justify-between text-sm">
```
ItsANameToo, line 24: would be good if the cursor changes to `not-allowed` when it's disabled, currently it still shows cursor-pointer instance while you can't interact with it

resources/js/Pages/Collections/Components/CollectionsFilter/CollectionsNetworksFilter.tsx
```diff
@@ -0,0 +1,83 @@
...
+    const { t } = useTranslation();
+
+    return (
+        <div className="flex flex-col overflow-hidden rounded-xl bg-white shadow-xl lg:max-h-screen">
+            <div className="bg-theme-secondary-50 px-6 py-3 text-sm font-medium text-theme-secondary-700">
+                {t("pages.collections.collections_network")}
+            </div>
+
+            <div className="flex flex-col gap-3 px-6 pb-6 pt-3">
+                {availableNetworks.map((network) => (
```
ItsANameToo, line 69: can you check if for the testnet versions you can show `Mumbai` and `Goerli` respectively instead of duplicating polygon/ethereum for them?

ItsANameToo (changes requested):

ItsANameToo (changes requested):
a couple more issues with this PR to resolve:
- when a network filter is active, it would be good to indicate that on the filter like we do when the "show hidden" toggle is active. This will highlight to the user that they have some filter active to avoid confusion
image

- if you hide a polygon NFT and filter by ETH and then switch to "show hidden", it will disable the network filter in the UI (ETH is no longer selected), but it won't actually stop filtering, causing the below scenario:
image

found two other issues but those are unrelated to the PR, so will add cards for them instead

ItsANameToo (changes requested):
still have some issues with the filter notification:
- if you only own NFTs for one network, it will indicate it's filtered even though it technically shows all your NFTs
image

- when you have all networks selected and toggle "show hidden", then deselect all networks and untoggle "show hidden" you are now in a state where you can't get rid of the nofitication dot anymore
image

- when the page loads the NFTs, it briefly shows the filter as active before it removes the notification dot. Mainly noticeable when the user owns a lot of NFTs or if the user has no NFTs at all
image

resources/js/Pages/Collections/Components/CollectionsFilter/CollectionsFilterPopover.tsx
```diff
@@ -45,8 +45,12 @@ export const CollectionsFilterPopover = ({
         }
     }, [debouncedQuery]);
 
-    const allNetworksSelected =
-        availableNetworks.filter((network) => network.collectionsCount > 0).length === selectedChainIds.length;
+    const isNetowrkFilterActive =
+        isLoading ||
```
ItsANameToo, line 49: the downside of this check is that you click a network and it will show the dot for a brief moment before ending up in the loading state where it gets hidden only to show again afterwards, resulting in a flickering when changing filters

resources/js/Pages/Collections/Components/CollectionsFilter/CollectionsFilterPopover.tsx
```diff
@@ -38,6 +45,13 @@ export const CollectionsFilterPopover = ({
         }
     }, [debouncedQuery]);
 
+    const isNetowrkFilterActive =
```
ItsANameToo, line 48:
```suggestion
    const isNetworkFilterActive =
```

resources/js/Pages/Collections/Components/CollectionsFilter/CollectionsFilterPopover.tsx
```diff
@@ -38,6 +45,13 @@ export const CollectionsFilterPopover = ({
...
+        isLoading ||
+        availableNetworks
+            .filter((network) => network.collectionsCount > 0)
+            .every((network) => selectedChainIds.includes(network.chainId)) ||
+        selectedChainIds.length === 0;
```
ItsANameToo, line 53: logic is still flawed when you only have one of the two networks available, as it will show the notification dot when "all" are deselected now

image

ItsANameToo (approved):

crnkovic (approved):

PR #82, opened 2023-09-12
fix: increment default size for collection banners

patricio0312rev:
# [[collections] Collection banner quality is low cause of width=500 px default response](https://app.clickup.com/t/861nbbvvj)

## Summary

- The default size for collection banners has been set from `500px` (coming in default response by Alchemy) to `1378px` (size we use for `FetchCollectionBanner` job).

## Steps to reproduce

1. Run the app in `test_e2e` mode, make sure you use a wallet with multiple NFTs from different collections.
2. Make sure you also run `php artisan horizon`.
3. Go to `/collections` page and wait a few seconds until the app fetches new collections from your connected wallet.
4. Click on any recently fetched collection and see the magic ✨ 

image

## Checklist

-   [ ] Ready to be merged


app/Http/Client/Alchemy/AlchemyPendingRequest.php
```diff
@@ -321,6 +321,13 @@ public function parseNft(array $nft, int $networkId): Web3NftData
 
         $mintTimestamp = $this->getNftMintingDateProperty($nft);
 
+        $bannerImageUrl = Arr::get($nft, 'contractMetadata.openSea.bannerImageUrl');
+        if (! empty($bannerImageUrl)) {
+            $bannerImageUrl = preg_replace('/(?<=\?|&)w=(\d+)/', 'w=1378', $bannerImageUrl);
```
ItsANameToo, line 326: why do the replacement here when there is one handling it in `NftImageUrl`? 🤔 I'd prefer if that one gets updated to match the available options from OpenSea so it's all in one place and that passes a size as well as opposed to changing it to a hardcoded value

ItsANameToo (changes requested):

app/Http/Client/Alchemy/AlchemyPendingRequest.php
```diff
@@ -321,6 +322,13 @@ public function parseNft(array $nft, int $networkId): Web3NftData
 
         $mintTimestamp = $this->getNftMintingDateProperty($nft);
 
+        $bannerImageUrl = Arr::get($nft, 'contractMetadata.openSea.bannerImageUrl');
+        if (! empty($bannerImageUrl)) {
+            $bannerImageUrl = NftImageUrl::get($bannerImageUrl, ImageSize::Banner);
+        } else {
+            $bannerImageUrl = null;
+        }
```
ItsANameToo, line 330: We already have banners that make use of 1378 as image size before your changes, so how did those get used (see clickup for examples)?  or were those fetched by mnemonic? if so, might be worth it to change it there as well (+ moralis) for consistency so everything gets updated with similar banner sizes. Right now this solution solely applies to alchemy while there are other providers that can be used for banners as well

ItsANameToo (changes requested):

ItsANameToo (approved):

PR #83, opened 2023-09-13
fix: improve types and handling for storing collection's banner_update_at

patricio0312rev:
# [[collections] error 500 when using banner](https://app.clickup.com/t/861nbgpah)

## Summary

https://app.clickup.com/t/861nbgpah

## Checklist

-   [ ] Ready to be merged

### DIFF
--- a/app/Jobs/FetchCollectionBanner.php
+++ b/app/Jobs/FetchCollectionBanner.php
@@ -41,7 +41,7 @@ class FetchCollectionBanner implements ShouldBeUnique, ShouldQueue
         );
 
         $this->collection->extra_attributes->set('banner', $banner);
-        $this->collection->extra_attributes->set('banner_updated_at', now()->timestamp);
+        $this->collection->extra_attributes->set('banner_updated_at', now()->toIso8601String());
 
         $this->collection->save();
     }
--- a/app/Jobs/FetchCollectionBanner.php
+++ b/app/Jobs/FetchCollectionBanner.php
@@ -41,7 +41,7 @@ class FetchCollectionBanner implements ShouldBeUnique, ShouldQueue
         );
 
         $this->collection->extra_attributes->set('banner', $banner);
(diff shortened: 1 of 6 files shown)


app/Models/Collection.php
```diff
@@ -117,6 +117,10 @@ public function bannerUpdatedAt(): ?Carbon
     {
         $bannerUpdatedAt = $this->extra_attributes->get('banner_updated_at');
 
+        if ($bannerUpdatedAt === null) {
+            return null;
+        }
+
         return Carbon::parse($bannerUpdatedAt);
```
ItsANameToo, line 124: remove parsing, let the callee handle that and just return the value as-is from `$this->extra_attributes->get('banner_updated_at');`

ItsANameToo (changes requested):

app/Jobs/FetchCollectionBanner.php
```diff
@@ -41,7 +41,7 @@ public function handle(): void
         );
 
         $this->collection->extra_attributes->set('banner', $banner);
-        $this->collection->extra_attributes->set('banner_updated_at', now()->timestamp);
+        $this->collection->extra_attributes->set('banner_updated_at', now()->toIso8601String());
```
ItsANameToo, line 44: you can keep the timestamp, but in that case change the `CollectionDetailData` signature so it matches the actual data being stored

ItsANameToo: advantage of timestamp is that you can change your check in the collection controller from `$bannerUpdatedAt->diffInDays(now()) > 7` to `$bannerUpdatedAt < now()->subDays(7)->timestamp` so you don't have to parse it

ItsANameToo: in that case I'd simply use `Carbon::now()` as that already stores a string representation and will keep it in line with other `something_at` fields we use (e.g. see `Jobs/FetchNativeBalances.php`. In that case also change the other instance where you added `toIso8601String` in a prior PR

PR #86, opened 2023-09-13
refactor: batch native balance requests

shahin-hq:
## Summary
Closes: https://app.clickup.com/t/861nbbmrx

## Checklist

-   [x] Ready to be merged


app/Console/Commands/FetchNativeBalances.php
```diff
@@ -56,18 +57,22 @@ private function handleAllWallets(Collection $networks): void
         $onlyOnline = $this->option('only-online');
 
         Wallet::query()
+            ->select(['id', 'address'])
             ->when($onlyOnline, fn ($query) => $query->online())
             ->when(! $onlyOnline, fn ($query) => $query->recentlyActive())
             ->chunkById(100, function ($wallets) use ($networks) {
-                $wallets->each(fn ($wallet) => $this->handleWallet($wallet, $networks));
+                $this->process($wallets, $networks);
```
oXtxNt9U, line 64: We process wallets in chunks of `100`, while the Moralis endpoint only accepts up to `25` addresses.

This will probably stop working once we exceed the threshold (unless I missed something). So maybe we just chunk by 25 here?

https://docs.moralis.io/web3-data-api/evm/reference/get-native-balances-for-addresses?chain=eth&wallet_addresses=[]

image

oXtxNt9U (changes requested):

app/Console/Commands/FetchNativeBalances.php
shahin-hq: Ah, right! Nice catch. Should be fine now.

app/Jobs/FetchNativeBalances.php
```diff
@@ -6,52 +6,102 @@
...
+                $balancesToInsert->toArray(),
+                ['wallet_id', 'token_id'],
+                ['balance', 'updated_at']
+            );
+        });
+
+        $walletsToUpdate->map(function ($wallet) {
+            $wallet->extra_attributes->set('native_balances_fetched_at', Carbon::now());
+            $wallet->save();
         });
```
crnkovic, line 95: Is there a reason this isn't a part of the database transaction?

app/Jobs/FetchNativeBalances.php
```diff
@@ -6,52 +6,102 @@
...
+        $balances = Moralis::getNativeBalances($addresses, $this->network);
+
+        $walletsToUpdate = collect();
+
+        $balancesToInsert = $balances->map(function ($walletBalance) use ($nativeToken, $walletsToUpdate) {
+            $address = Str::lower($walletBalance->address);
+
+            $wallet = $this->wallets->first(fn ($wallet) => $address === Str::lower($wallet->address));
+
+            $walletsToUpdate->push($wallet);
```
crnkovic, line 64: Won't this `$walletsToUpdate` variable always match `$this->wallets`? You don't need to hand-pick which wallets are going to be updated, you can just do `$this->wallets->map()` to save their `native_balances_fetched_at`

app/Jobs/FetchNativeBalances.php
crnkovic: Also, this is precisely why I hate these JSON columns, they just make things difficult, with no real benefits 🤷  I'd be totally okay if you just added the new timestamp column and then you could update in one query 🙈

shahin-hq: I didn't put too much though on moving `native_balances_fetched_at` into transaction, however, I should, if balance update fails then  `native_balances_fetched_at` should have `null`/old value, so the job can pick it next time.

shahin-hq: I checked many places to find at what point it gets `[]` set but couldn't find it, if I could fix that I'd set `native_balances_fetched_at` in one query..

app/Jobs/FetchNativeBalances.php
shahin-hq: There is no guarantee that API will return balances for all provided wallets, that is why I chose to be on the safe side. If you are 100% sure, I can remove that part.

crnkovic: Even with randomly made up addresses, they still return 0 for balance (see image). Doesn't really matter as it will query 25 wallets, whatever you feel like

Screenshot 2023-09-15 at 12 05 25

shahin-hq: Thanks for checking. Changed it as you suggested

app/Console/Commands/FetchNativeBalances.php
```diff
@@ -56,18 +57,22 @@ private function handleAllWallets(Collection $networks): void
         $onlyOnline = $this->option('only-online');
 
         Wallet::query()
+            ->select(['id', 'address'])
             ->when($onlyOnline, fn ($query) => $query->online())
             ->when(! $onlyOnline, fn ($query) => $query->recentlyActive())
-            ->chunkById(100, function ($wallets) use ($networks) {
-                $wallets->each(fn ($wallet) => $this->handleWallet($wallet, $networks));
+            ->chunkById(25, function ($wallets) use ($networks) {
```
ItsANameToo, line 63: maybe useful to make this into a constant and add the reasoning for using 25 here, otherwise someone might change that without knowing why it was done as we usually chunk by 100

shahin-hq: Fair enough. Fixed.

ItsANameToo (approved):

crnkovic (approved):

PR #89, opened 2023-09-14
refactor: clean up the NFT handler

crnkovic:
## Summary

https://app.clickup.com/t/862kdkkjh

This PR does a bit of housekeeping for our `Web3NftHandler` class -- a class that takes the array of NFTs from the provider and ensures they are added to the database, the matching collection is created, traits are created, and jobs are dispatched.

First, since we will be working with Collection models, I have refactored the `DetermineCollectionMintingDate` job class to use the model.

Next, we can fine-tune whether to dispatch the `FetchCollectionFloorPrice` jobs when new collections are created (and have no floor price). We always do this, so I just removed this `$dispatchJobs` argument as it will always do that.

After all the records in the database are created, we update the collection's fiat value and the user's total collection value. Previously, this has always run, even if the response returned no NFTs. Now, I added the check to ignore the entire class if there are no NFTs, saving us a few SQL queries.

The big difference in this PR is that [raw SQL query](https://github.com/ArdentHQ/dashbrd/pull/89/files#diff-1831d72a37ff7448c5c69621035d341fd3e45b76ed181e5c021cbd9d35efcf45L92-L107) to UPSERT collections has been replaced with `Collection::upsertReturning()` call [here](https://github.com/ArdentHQ/dashbrd/pull/89/files#diff-1831d72a37ff7448c5c69621035d341fd3e45b76ed181e5c021cbd9d35efcf45R68-R70). This allows us to remove stuff like [this](https://github.com/ArdentHQ/dashbrd/pull/89/files#diff-1831d72a37ff7448c5c69621035d341fd3e45b76ed181e5c021cbd9d35efcf45L81). This method replicates the raw SQL query (`UPSERT RETURNING`), but also hydrates the Collection models, so we can use them later on.

There are no feature changes to this PR, it will still do the same thing, it's just that the class is now much easier to work with and to maintain.

When reviewing, open the new class and the old class side by side as the diff is a bit hard to see.

## Checklist

-   [ ] Ready to be merged


app/Support/Web3NftHandler.php
```diff
@@ -39,149 +40,134 @@ public function __construct(
...
 
-        $ids = DB::transaction(function () use ($nfts, $collectionsData, $valuesPlaceholders, $dispatchJobs, $now) {
-            // upsert nfts/collections (if any)
-            if ($nfts->isEmpty()) {
-                return collect();
-            }
+        // We take all the columns (except the ones we unique on which are network_id and address)
+        // and generate `column = coalesce(excluded.column, collections.column)` for UPSERT...
+        $update = collect($columns)->mapWithKeys(fn ($column) => [
+            $column => DB::raw(sprintf('coalesce(excluded.%s, collections.%s)', $column, $column)),
```
oXtxNt9U, line 62: A few differences compared to previous version:

- we no longer trim the name, not sure if that's needed though

```
name = trim(coalesce(excluded.name, collections.name)),
```

- now we always keep a non-null floor price, while previously we'd potentially null it:

```
floor_price_token_id = excluded.floor_price_token_id,
floor_price_retrieved_at = excluded.floor_price_retrieved_at,
```

- this situation sounds similar to https://github.com/ArdentHQ/dashbrd/pull/98#discussion_r1328635128

so maybe we just never update it if it's already set ? 

i.e. something like 
`floor_price_token_id = coalesce(collections.floor_price_token_id, excluded.floor_price_token_id)`

app/Support/Web3NftHandler.php
```diff
@@ -39,149 +40,134 @@ public function __construct(
...
-            $groupedByAddress = collect($dbCollections)->keyBy(fn ($nft) => Str::lower($nft->address));
-
-            $nfts = $nfts->filter(function ($nft) use ($groupedByAddress) {
-                $collection = $groupedByAddress->get(Str::lower($nft->tokenAddress));
-
-                return $this->shouldKeepNft($collection);
-            });
+        $collections = DB::transaction(function () use ($nfts, $collectionsData, $update, $now) {
+            /** @var EloquentCollection<int, CollectionModel> $collections */
+            $collections = CollectionModel::upsertReturning(
```
oXtxNt9U, line 67: Note to self,  `upsertReturning` comes from [here](https://github.com/tpetry/laravel-postgresql-enhanced/blob/b89af5de3ee36ee236525dc207f452629211c0fd/src/Eloquent/Mixins/BuilderReturning.php#L106)

oXtxNt9U (changes requested):

app/Support/Web3NftHandler.php
crnkovic: Regarding trimming, meh, we can just trim on the PHP level.

For the floor price being null, I'm kinda curious - why did we even null the price before? If the price is already set in the response, should we just set it in the database?

oXtxNt9U: > For the floor price being null, I'm kinda curious - why did we even null the price before? If the price is already set in the response, should we just set it in the database?

Likely just an oversight. It might even have been there before we added the floor price job for Mnemonic. But we should change it now while at it for consistency.

oXtxNt9U: > Regarding trimming, meh, we can just trim on the PHP level.

:+1:

crnkovic: Hmm, with my PR, we already do take the first non-null value. Though, I suspect the order matters, right?

If both are set, will take from the database table:
```
floor_price_token_id = coalesce(collections.floor_price_token_id, excluded.floor_price_token_id)
```

If both are set, will take from the array to update:
```
floor_price_token_id = coalesce(excluded.floor_price_token_id, collections.floor_price_token_id)
```

oXtxNt9U: >  Though, I suspect the order matters, right?

Yes, we should use the former variant for this column. It will ensure we only write it if we don't have one yet, then it's consistent with https://github.com/ArdentHQ/dashbrd/pull/98#discussion_r1328635128 and only the actual floor price job is responsible to update the price afterwards.

Having an initial price means we have more likely something to show in the UI before the floor price job ran. That's the only reason why we should bother setting a price here.

oXtxNt9U (changes requested):
`composer db:live` fails for me now:

image

Maybe also good to sanity check that `LiveDumpNfts` and `LiveDumpWallets` still work just in case they are affected by any of the changes.

oXtxNt9U (approved):

ItsANameToo:
what's remaining for this?

crnkovic:
many PRs have been merged in between so I have to resolve conflicts and test that it still works as expected

ItsANameToo:
Conflicts @crnkovic

crnkovic:
With some of the changes I introduced in one of the [this](https://github.com/ArdentHQ/dashbrd/commit/5ef126fc090d74daf1a1c1a43d91b0f8e431512b) PR, this has become a mess. The library we used to limit the eager loaded relationship (`staudenmeir/eloquent-eager-limit`) plugs into Laravel's Query builder... but so does `tpetry/laravel-postgresql-enhanced`, that we also use.

This creates a conflict as one library doesn't know about the other, so we'd have to override the [query builder and grammar](https://github.com/ArdentHQ/dashbrd/pull/89/files#diff-2dbe9cf1a0416f82f62e51746210e5ffb649ff95c87c51f9dffd4eea5324b0c7) to make them aware of each other and is this is just a mess.

Closing this for now until I figure out the better solution, but leaving the branch as I still believe these changes will clean up the codebase.

PR #101, opened 2023-09-18
refactor: do not index NFTs for potentially "full" collections

crnkovic:
## Summary

https://app.clickup.com/t/861nbbgp1

## Checklist

-   [ ] Ready to be merged

### DIFF
--- a/app/Models/Collection.php
+++ b/app/Models/Collection.php
@@ -417,4 +417,22 @@ class Collection extends Model
 
         return $this->last_viewed_at->gte(now()->subDays(1));
     }
+
+    /**
+     * A collection is only *potentially* full if the total supply matches the last token number that was indexed for the collection.
+     *
+     * Or if the number of NFTs in a collection matches the total supply.
+     */
+    public function isPotentiallyFull(): bool
+    {
+        if ($this->last_indexed_token_number === null || $this->supply === null) {
+            return false;
+        }
+
+        if ((string) $this->supply === (string) $this->last_indexed_token_number) {
+            return true;
+        }
+
+        return $this->supply === $this->nfts()->count();
+    }
 }
--- a/app/Console/Commands/FetchCollectionNfts.php
+++ b/app/Console/Commands/FetchCollectionNfts.php
@@ -31,7 +31,11 @@ class FetchCollectionNfts extends Command
     public function handle(): int
     {
         $this->forEachCollection(function ($collection) {
-            Job::dispatch($collection, $this->option('start-token') ?? $collection->last_indexed_token_number);
+            Job::dispatch(
+                $collection,
+                startToken: $this->option('start-token') ?? $collection->last_indexed_token_number,
+                skipIfPotentiallyFull: true,
+            );
         });
 
         return Command::SUCCESS;
--- a/app/Jobs/FetchCollectionNfts.php
+++ b/app/Jobs/FetchCollectionNfts.php
@@ -28,7 +28,8 @@ class FetchCollectionNfts implements ShouldBeUnique, ShouldQueue
      */
     public function __construct(
         public Collection $collection,
-        public ?string $startToken = null
+        public ?string $startToken = null,
+        public bool $skipIfPotentiallyFull = false,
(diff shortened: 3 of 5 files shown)


app/Models/Collection.php
```diff
@@ -417,4 +417,22 @@ public function recentlyViewed(): bool
...
+     *
+     * Or if the number of NFTs in a collection matches the total supply.
+     */
+    public function isPotentiallyFull(): bool
+    {
+        if ($this->last_indexed_token_number === null || $this->supply === null) {
+            return false;
+        }
+
+        if ((string) $this->supply === (string) $this->last_indexed_token_number) {
```
ItsANameToo, line 432: this is a tricky check, because the token numbers often (not always) start from 0, causing a one-off issue here. There are also various NFTs that don't have a logical numbering for their tokens, for example `AlphaDogs` (address: `0x32f8EE2B5707138e1bdd04D3631A04EB104dc141`, supply is 4625 but their token numbers are strings like `77699296229261824`)

app/Models/Collection.php
```diff
@@ -417,4 +417,22 @@ public function recentlyViewed(): bool
...
+            return true;
+        }
+
+        return $this->supply === $this->nfts()->count();
```
ItsANameToo, line 436: this is also an interesting one as the count can be more than the supply as well, meaning that we indexed more than the contract indicates there should be (it doesn't happen that often, but it's for example the case for `Cool Cats Fractures` where the database count is 11k and the supply is 5986

ItsANameToo (changes requested):

app/Models/Collection.php
ItsANameToo: logic wise maybe the following:
- if supply === last token number => full (handles general case where 10k supply results in 10k token numbers)
- if supply === last token number - 1 && db count === supply => full (handles cases where token number starts from 0)
- if supply < last token number && db count === supply => full (handles cases where token numbers are arbitrary values, usually much larger values than the supply is)
- if supply > db count => not full (for cases where we have supply < count (e.g. burn metrics)). The assumption we will make here is that we can handle the burn scenario by returning `full` by default if this last check does not match, as they generally start with their full collection at first (e.g. cool cats fractures at 11.5k) and burn over time, causing the supply to decrease over time while we already indexed them all

ItsANameToo: @crnkovic ^

alfonsobries (approved):

app/Models/Collection.php
```diff
@@ -479,6 +479,49 @@ public function recentlyViewed(): bool
...
     }
 
+    /**
+     * Determine whether the collection is *potentially* full. We call it "potentially" because there is no way for
+     * us to be 100% sure that the collection is full and we make our best guess.
+     */
+    public function isPotentiallyFull(): bool
+    {
+        // If there is no supply or if the NFTs for the collection haven't been indexed, we know for sure it's not full...
+        if ($this->last_indexed_token_number === null || $this->supply === null) {
```
shahin-hq, line 489: Please note that we are already filtering out collections with `supply=null`: https://app.clickup.com/t/85zu0hdha

It's worth double-checking to ensure that we are not excessively filtering out collections.

https://github.com/ArdentHQ/dashbrd/pull/101/files#diff-5254893bd34d47c136df16ce1c571ddd8d62d0bd9a0d1730edecbbb049fa7d89R52

crnkovic: wdym? we're filtering collections where the supply is not null so this check (`$this->supply === null`) will always be false and therefore irrelevant in this case

alfonsobries (approved):

app/Models/Collection.php
ItsANameToo: with the current way we call this `isPotentiallyFull` method we won't ever pass it a collection with a `null` supply, but for reliability I'd keep the check in case something calls it without the `withAcceptableSupply` scope as it avoids assumptions here

ItsANameToo (approved):

PR #103, opened 2023-09-18
feat: log relevant activities

alfonsobries:
## Summary

Adds logs for commands

Closes https://app.clickup.com/t/861nannku

## Checklist

-   [ ] Ready to be merged


ItsANameToo (changes requested):
in general I'm not sure if we need these logged, because:
- the queued jobs show up in horizon already, including the models they apply to
- they don't add very interesting details, as it's mainly listing the models that the command will cover
- with https://github.com/ArdentHQ/dashbrd/pull/108 it already adds additional logging for the individual models, which makes the general lists redundant as it will result for example in a list of addresses for collections that are being used and that will be followed by the individual collections with before/after data from https://github.com/ArdentHQ/dashbrd/pull/108, making the list unnecessary imo

crnkovic:
I agree with @ItsANameToo. Logging when jobs are dispatched seems redundant, we just have to trust our database that it will fetch proper records. Also, Horizon supports [tagging](https://laravel.com/docs/10.x/horizon#tags) which we can use to see which jobs ran for which models.

When it comes to logging, I think "edge cases" are much more interesting, stuff like UPSERTs to the database, early return cases such as [this](https://github.com/ArdentHQ/dashbrd/blob/develop/app/Jobs/FetchNativeBalances.php#L71-L73), removing data from database such as [here](https://github.com/ArdentHQ/dashbrd/blob/develop/app/Jobs/FetchTokens.php#L84-L92) (log IDs of models removed), etc.

Just like I commented on the ClickUp card, I feel logging requests and responses to external services would give us more context when debugging than randomly calling `Log::info()` all over the codebase.

PR #108, opened 2023-09-19
feat: log relevant activities for jobs

alfonsobries:
## Summary

Did my best to add what I think is relevant data, we can add more data or logs (or remove) as needed

https://app.clickup.com/t/861nannku

## Checklist

-   [ ] Ready to be merged


app/Jobs/FetchCollectionNfts.php
```diff
@@ -38,13 +39,31 @@ public function __construct(
...
+                'collection' => $this->collection->address,
+                'startToken' => $this->startToken,
+            ]);
+
             return;
         }
 
         // Ignore collections above the supply cap
         if ($this->collection->supply === null || $this->collection->supply > config('dashbrd.collections_max_cap')) {
+            Log::info('FetchCollectionNfts Job: Ignored becasue supply is null or > max', [
```
ItsANameToo, line 60:
```suggestion
            Log::info('FetchCollectionNfts Job: Ignored because supply is null or > max', [
```

app/Jobs/FetchEnsDetails.php
```diff
@@ -35,6 +36,10 @@ public function __construct(
      */
     public function handle(): void
     {
+        Log::info('FetchEnsDetails Job: Processing', [
+            'wallet' => $this->wallet->address,
```
ItsANameToo, line 40: also log the ENS name here and at the end so we can see differences between the two in case they changed it

app/Jobs/FetchNftActivity.php
```diff
@@ -42,9 +43,19 @@ public function uniqueId(): string
      */
     public function handle(): void
     {
+        Log::info('FetchNftActivity Job: Processing', [
+            'token_number' => $this->nft->token_number,
```
ItsANameToo, line 47: best to include the collection id or name as well, otherwise these token numbers will be ambiguous between collections

app/Jobs/FetchNftActivity.php
```diff
@@ -89,6 +100,14 @@ public function handle(): void
             FetchNftActivity::dispatch($this->nft)->onQueue(Queues::SCHEDULED_WALLET_NFTS);
         }
 
+        Log::info('FetchNftActivity Job: Handled', [
+            'address' => $collection->address,
+            'network' => $collection->network->id,
+            'token_number' => $this->nft->token_number,
+            'upserted_activities' => $upserted,
```
ItsANameToo, line 107: maybe include a count only, otherwise there will be lots of activities logged which we can view in the db already 🤔

app/Jobs/FetchPriceHistory.php
```diff
@@ -38,7 +39,19 @@ public function __construct(
      */
     public function handle(MarketDataProvider $provider): void
     {
+        Log::info('FetchPriceHistory Job: Processing', [
```
ItsANameToo, line 42: maybe add token name as well, as that's easy to recognise

ItsANameToo: applies to all these logs in this file

app/Jobs/FetchWalletNfts.php
```diff
@@ -65,6 +73,15 @@ public function handle(): void
         GalleryCache::clearAllDirty();
 
         UserCache::clearAll($this->wallet->user);
+
```
ItsANameToo, line 76: a few lines above there is `$nftHandler->cleanupNftsAndGalleries($this->startTimestamp);`, maybe also add logs in there when a gallery is adjusted? that would highlight any changes that are being made to them from jobs

ItsANameToo (changes requested):

alfonsobries:
@ItsANameToo feedback handled

app/Jobs/FetchWalletNfts.php
```diff
@@ -58,13 +66,30 @@ public function handle(): void
         );
 
         if ($result->nextToken === null) {
+            Log::info('FetchWalletNfts Job: run cleanupNftsAndGalleries', [
+                'wallet' => $this->wallet->address,
+                'network' => $this->network->id,
+                'next_token' => $result->nextToken,
```
ItsANameToo, line 72: will always be `null` as you're inside the if case

alfonsobries: removed

ItsANameToo (approved):

PR #116, opened 2023-09-20
feat: add article model

shahin-hq:
## Summary
Closes: https://app.clickup.com/t/8678v08n5

> We will require a new article model with the following properties
> 
> title - string
> category - string, will be an enum in our code
> date - nullable date
> meta_description - nullable text, should have enough room for a description for SEO purposes
> content - text, actual content of the article
> author - a single user who wrote the article
> add softdeletes for articles so editors can delete them without it being completely gone
> add views similar to how we track gallery views
> cover image through media library
> references collections (1 article will have between 1 and 8 collections to feature)

## Checklist

-   [ ] Ready to be merged


ItsANameToo:
@shahin-hq point it to `feat/articles` instead of develop

shahin-hq:
@ItsANameToo please update the `feat/articles` branch before reviewing.

app/Models/Article.php
```diff
@@ -0,0 +1,36 @@
...
+use Spatie\MediaLibrary\HasMedia;
+use Spatie\MediaLibrary\InteractsWithMedia;
+
+class Article extends Model implements HasMedia, Viewable
+{
+    use BelongsToUser, CanBeLiked, HasFactory, InteractsWithMedia, InteractsWithViews, SoftDeletes;
+
+    public $guarded = ['id'];
+
+    protected function likesTable(): string
```
ItsANameToo, line 24: we don't need to track likes for articles, just views

database/factories/ArticleFactory.php
```diff
@@ -0,0 +1,35 @@
...
+     */
+    public function definition(): array
+    {
+        return [
+            'title' => fake()->name(),
+            'category' => fake()->name(),
+            'published_at' => fake()->date(),
+            'meta_description' => fake()->text(),
+            'content' => fake()->text(),
+            'user_id' => User::factory()->withWallet(),
```
ItsANameToo, line 32:
```suggestion
            'user_id' => fn () => User::factory()->withWallet(),
```

so we can pass it a user as well

ItsANameToo (changes requested):

database/factories/ArticleFactory.php
crnkovic: think you can still override the `user_id`, if you pass the factory instance it will lazily create it only if you don't manually set `user_id`

ItsANameToo: ah indeed, I was under the impression it required `fn()` for that

ItsANameToo (approved):

PR #124, opened 2023-09-21
fix: condition to check for metamask existence

alfonsobries:
## Summary

Issue is cause window.ethereum is assigned a html collection (apparently because the id in a svg), updating for a more error-proof condition which is enough to fix the issue but also removed the id on the svgs

image

Closes https://app.clickup.com/t/862kdwykc

## Checklist

-   [ ] Ready to be merged

### DIFF
--- a/resources/js/Hooks/useMetaMask.ts
+++ b/resources/js/Hooks/useMetaMask.ts
@@ -62,8 +62,7 @@ const networkConfigs: Record<number, object> = {
     },
 };
 
-// eslint-disable-next-line @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-explicit-any
-const hasMetaMask = (): boolean => window.ethereum?.isMetaMask ?? false
+const hasMetaMask = (): boolean => window.ethereum?.isMetaMask ?? false;
 
 const getEthereum = (): Ethereum => window.ethereum as Ethereum;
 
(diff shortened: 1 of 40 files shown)


resources/js/Hooks/useMetaMask.ts
```diff
@@ -62,7 +62,8 @@ const networkConfigs: Record<number, object> = {
     },
 };
 
-const hasMetaMask = (): boolean => window.ethereum !== undefined;
+// eslint-disable-next-line @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-explicit-any
```
alfonsobries, line 65: Note: using explicit `any` locally is not necessary since it properly handles the ethereum type that is defined on the `resources/types/globals.d.ts` but for some reason does not work here and creates a linting issue

resources/js/Hooks/useMetaMask.ts
```diff
@@ -62,7 +62,8 @@ const networkConfigs: Record<number, object> = {
...
+const hasMetaMask = (): boolean => window.ethereum?.isMetaMask === true;
```
ItsANameToo, line 66:
```suggestion
const hasMetaMask = (): boolean => window.ethereum?.isMetaMask ?? false
```
try this? then it defaults to a boolean still in case window.ethereum is not defined 🤔

ItsANameToo: causes other lint issues, reverting

ItsANameToo (approved):

PR #125, opened 2023-09-21
refactor: limit fetched collections for signed wallets

alfonsobries:
## Summary

Closes part of https://app.clickup.com/t/8678vbt3e

## Checklist

-   [ ] Ready to be merged


app/Models/Collection.php
```diff
@@ -466,4 +466,13 @@ public function isBlacklisted(): bool
...
         return BlacklistedCollections::includes($this->address);
     }
+
+    /**
+     * @param  Builder<self>  $query
+     * @return Builder<self>
+     */
+    public function scopeWithSignedWallet(Builder $query): Builder
+    {
+        return $query->whereHas('nfts', fn ($nftQuery) => $nftQuery->whereHas('wallet', fn ($walletQuery) => $walletQuery->hasBeenSigned()));
```
oXtxNt9U, line 476: Hmm, for very large collections this might be very slow. Worst case we have to scan each nft in each collection.

Maybe alternatively we can use a special purpose query.

For example, something like this:

```sql
with signed_wallets as (
    select wallet_id from wallets
    -- should be indexed, even better if it has a separate column because it's simpler to follow
    where extra_attributes->last_signed_at is not null 
), collection_ids as (
    select distinct collection_id from nfts
    where nfts.wallet_id in (select wallet_id from signed_wallets)
)
select * from collections
where in in (select collection_id from collection_ids)
order by ...
limit ...
```

(haven't tested it, but you get the idea)

oXtxNt9U (changes requested):
Left one comment regarding the query to scope collections to signed wallets.

app/Models/Collection.php
ItsANameToo: see discussion on slack @alfonsobries

alfonsobries:
@ItsANameToo @oXtxNt9U updated so it includes the last_signed_at date as a column instead of a extra attribute and adds the index

https://github.com/ArdentHQ/dashbrd/pull/125/files/fdf40d22c11f0246388b39ea7540cc2ffb8e4b28..df6d32c11c7040b08ad270cba126953e62f5a4a3

oXtxNt9U (approved):

samharperpittam (approved):

ItsANameToo (approved):

PR #126, opened 2023-09-22
feat: fetch collection nfts limited for non signed wallets

alfonsobries:
## Summary

Depends on https://github.com/ArdentHQ/dashbrd/pull/125

Closes point 2 from https://app.clickup.com/t/8678vbt3e

schedules the job for fetching nft collection for all collections with a limit, but first it fetches the ones that were not fetched yet or the oldest ones

## Checklist

-   [ ] Ready to be merged


ItsANameToo:
some conflicts now @alfonsobries

oXtxNt9U (approved):

app/Console/Commands/FetchCollectionNfts.php
```diff
@@ -33,25 +32,26 @@ class FetchCollectionNfts extends Command
     public function handle(): int
     {
         $onlySigned = (bool) $this->option('only-signed');
+        $limit = $this->option('limit');
 
         if ($onlySigned) {
             Collection::getWithSignedWallet()->each(function (Collection $collection) {
-                if (! $collection->isBlacklisted()) {
```
ItsANameToo, line 39: why is the blacklist check removed from here? 🤔

app/Console/Commands/FetchCollectionNfts.php
```diff
@@ -33,25 +32,26 @@ class FetchCollectionNfts extends Command
...
-                    );
-                }
+                FetchCollectionNftsJob::dispatch(
+                    $collection,
+                    $this->option('start-token') ?? $collection->last_indexed_token_number
+                );
             });
         } else {
-            $this->forEachCollection(function ($collection) {
-                if (! $collection->isBlacklisted()) {
```
ItsANameToo, line 48: likewise here

app/Console/Kernel.php
```diff
@@ -141,6 +141,17 @@ private function scheduleJobsForCollections(Schedule $schedule): void
             ->withoutOverlapping()
             ->dailyAt('5:30');
 
+        $schedule
+            ->command(FetchCollectionNfts::class, [
+                // Limit the number of NFTs to fetch for unsigned wallets, since
+                // it sorts the collections by last_fetched_at eventually it will
+                // fetch all NFTs for all collections
+                '--limit' => config('dashbrd.daily_max_nft_retrieval_for_unsigned_wallets'),
```
ItsANameToo, line 149: this limit was intended to be on the collections itself, not their NFTs. Idea is to take a subset of the new collections from unverified users and index those in full and pick up on other ones over the next few days. The issue with limiting it by NFTs is that it may take a very long time to complete it with the currently set limits (e.g. 100 NFTs would require 90 runs to complete for a collection with 9k NFTs)

ItsANameToo (changes requested):

app/Console/Commands/InteractsWithCollections.php
```diff
@@ -30,12 +30,21 @@ public function forEachCollection(Closure $callback, Closure $queryCallback = nu
...
-            }, 'collections.id', 'id');
+            ->when($limit !== null, fn ($query) => $query
+                ->limit($limit)
+                ->get()
+                ->filter(fn ($collection) => ! $collection->isBlacklisted())
+                ->each($callback)
+            )
+            ->when($limit == null, fn ($query) => $query->chunkById(100, function ($collections) use ($callback) {
+                $collections
+                    ->filter(fn ($collection) => ! $collection->isBlacklisted())
```
shahin-hq, line 45: This implementation breaks [these](https://app.clickup.com/t/85zu0hdha) remarks:

> spam => ignore for fetching NFTs and metadata / banner
> supply null/too large => ignore for fetching NFTs
> blacklist => ignore for fetching NFTs

The `InteractsWithCollections` is being used by many jobs and not all of them should filter blacklisted collections, since the default value of `limit=null` in the function signature this will cause unexpected results.

app/Console/Commands/FetchCollectionNfts.php
```diff
@@ -33,25 +32,26 @@ class FetchCollectionNfts extends Command
...
+            $this->forEachCollection(
+                callback: function ($collection) {
                     FetchCollectionNftsJob::dispatch(
                         $collection,
                         $this->option('start-token') ?? $collection->last_indexed_token_number
                     );
-                }
-            }, queryCallback: fn (Builder $query) => $query->withAcceptableSupply());
+                },
+                queryCallback: fn ($query) => $query->orderByOldestNftLastFetchedAt(),
```
shahin-hq, line 52: You need to keep the `supply` checks in place, otherwise it will try to pull NFTs collections with 20K supplu which will drain our API budget.

shahin-hq (changes requested):
Please take a look at the #139 before making any changes as it will probably make things easier for you. And I think `InteractsWithCollections->forEachCollections` is doing too much stuff, maybe it is wise to move some logic to another method.

app/Console/Kernel.php
alfonsobries: @ItsANameToo it limiting the collections, the setting name was incorrect, fixed on my next batch of changes

alfonsobries:
@ItsANameToo @shahin-hq should be fine now, let me know

tests/App/Models/CollectionTest.php
```diff
@@ -1052,3 +1052,42 @@
...
+    ]);
+
+    // never fetched
+    $collection5 = Collection::factory()->create();
+
+    $ids = Collection::orderByOldestNftLastFetchedAt()->pluck('id')->toArray();
+
+    expect($ids)->toEqual([
+        $collection3->id,
+        $collection5->id,
```
shahin-hq, line 1088: Are you sure that database will always return rows in this order: `collection3` then `collection5`?
Both of these values are `NULL` and we have this `orderBy` logic: `extra_attributes->>\'nft_last_fetched_at\' ASC NULLS FIRST` while  `toEqual` ensures order as well. This test looks like flaky to me..

ItsANameToo: what would you suggest instead for this test

shahin-hq: Removing `collection5` and having `collection3` would be sufficient as it will prove that `null` value comes before everything else.
I can make the change if you want to proceed.

ItsANameToo: please, then this can get merged

shahin-hq (approved):

ItsANameToo (approved):

PR #168, opened 2023-09-29
feat: fetch opensea slugs

alfonsobries:
## Summary

The job runs hourly but only fetches slugs for collections that were not fetched before and does not have the opensea slug already

Closes https://app.clickup.com/t/8678wqnc8

## Checklist

-   [ ] I checked my UI changes against the design and there are no notable differences
-   [ ] I checked my UI changes for any responsiveness issues
-   [ ] I checked my (code) changes for obvious issues, debug statements and commented code
-   [ ] I opened a corresponding card on Clickup for any remaining TODOs in my code
-   [ ] I added a short description on how to test this PR _(if necessary)_
-   [ ] I added a storybook entry for the component that was added _(if necessary)_
-   [ ] Documentation _(if necessary)_
-   [ ] Tests _(if necessary)_
-   [ ] Ready to be merged

PR #169, opened 2023-09-29
feat: rate limit opensea requests

alfonsobries:
## Summary

Closes https://app.clickup.com/t/8678wqn9v

See comments on pr

## Checklist

-   [ ] Ready to be merged


oXtxNt9U (approved):

ItsANameToo:
For numbers, locally I have 838 collections and only 95 were updated on this PR, while #168 updated 518. Granted that in reality we won't have that many collections at a time, it still shows that this ratelimit implementation has some issues to it

config/services.php
```diff
@@ -83,6 +83,14 @@
     'opensea' => [
         'key' => env('OPENSEA_API_KEY', null),
         'endpoint' => env('OPENSEA_API_ENDPOINT', 'https://api.opensea.io/api/'),
+
+        // @see https://docs.opensea.io/v1.0/reference/api-overview#api-faqs
+        'rate' => [
+            // Notice that the limit for POST request is 2, but currently we only
+            // use GET requests
+            'max_requests' => env('OPENSEA_MAX_REQUESTS', 4),
+            'per_seconds' => env('COPENSEA_MAX_REQUESTS_SECONDS', 1),
```
ItsANameToo, line 92:
```suggestion
            'per_seconds' => env('OPENSEA_MAX_REQUESTS_SECONDS', 1),
```

ItsANameToo (changes requested):

app/Jobs/FetchCollectionFloorPrice.php
```diff
@@ -100,4 +103,18 @@ public function retryUntil(): DateTime
...
+
+    /**
+     * Get the middleware the job should pass through.
+     *
+     * @return array<int, object>
+     */
+    public function middleware(): array
+    {
+        if (Config::get('dashbrd.web3_providers.'.self::class) === 'opensea') {
+            return [new RateLimited(Service::Opensea)];
```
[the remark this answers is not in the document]
ItsANameToo: tested this locally with the opensea request commented so it would only queue the job and update the collection's `opensea_slug_last_fetched_at` value. This caused no fails on #168 (838/838 collections updated), but on this PR it only updated around 10% (84/838)

alfonsobries:
@ItsANameToo updated

Issue was the same we use to have before we refactored most jobs to not depend on rate limit: Every time a job is rescheduled it counts as one retry or will be considered part of the total time required to run the job after the limit it will fail

Updated with the old solution: calculating the amount of time we need to run all jobs considering rate limit and set a `retryUntil` considering that limit

ItsANameToo:
possibly related discussion card: https://app.clickup.com/t/8678xagtr

alfonsobries:
@ItsANameToo @oXtxNt9U pr updated according to the card about spreading, think is a good solution and cleaner that the original solution I came up

oXtxNt9U (approved):

shahin-hq (approved):

ItsANameToo:
changing this to `develop` as it can be included in the next regular release

app/Console/Kernel.php
```diff
@@ -98,17 +100,34 @@ protected function schedule(Schedule $schedule): void
...
+
+        if (Config::get('dashbrd.web3_providers.'.FetchCollectionFloorPriceJob::class) === 'opensea') {
+            $schedule
+                ->command(FetchCollectionFloorPrice::class, [
+                    '--limit' => config('services.opensea.rate.max_requests'),
+                ])
+                ->withoutOverlapping()
+                // Opensea allows 4 requests per second, using 5 seconds to leave
+                // some room for other tasks or unexpected delays
+                ->everyFiveSeconds();
```
ItsANameToo, line 112: one downside of this is that we will forever queue a job with 5 second intervals to query the database to check if we need to update anything, and it will be no for the majority of the time. For example if we have updated all current collections, then we only want to update this hourly meaning the 5 second interval will run 720 queries before it will retrieve some actual collections to update (hour has passed at that point). At current this won't be that bad as the query is quick, but it's something we can improve upon

ItsANameToo: changing to draft as this bothers me still

app/Console/Commands/FetchCollectionOpenseaSlug.php
```diff
@@ -30,6 +30,8 @@ class FetchCollectionOpenseaSlug extends Command
      */
     public function handle(): int
     {
+        $limit = $this->option('limit');
```
ItsANameToo, line 33: I'd approach this differently. Instead of calling this command every 5 seconds and determining which collections have been updated since (and requiring a query for it), what about:
- keep the scheduled task on an hourly basis
- query one time for all collections that need updating
- then chunk those collections based on the allowed amounts for the ratelimit (e.g. 3)
- for each chunk, delay it by x seconds so they run after one another

That should handle the scenario in a similar way, but without the need to constantly query collections (every 5 seconds) when most collections won't require an update for the majority of the time. There are some caveats with this approach but they apply to the existing one as well, e.g. how we would handle failed jobs that get retried and clash with the newly scheduled ones. For this we should consider removing retries completely for certain jobs and let them fail and be re-scheduled later to not create additional jobs. Maybe we can already remove the retries for this job as it will be included in the next time the schedule gets triggered

@alfonsobries @oXtxNt9U

alfonsobries:
@ItsANameToo updated this query according to you proposal but with some adjustemnts

This what we should expect

- Both opensea jobs runs every hour
- It will calculate the number of jobs that use opensea and will use and the opensea limits to calculate how many jobs we can run in an hour and per time frame
- Since we have **2**jobs that use opensea, the limit is 4 per second means it will schedule (delay) 4 jobs every **2** seconds
- also added a threshold to prevent overlapping, means that every job is going to be delayed by a specific amount of seconds, We currently have to jobs so job 0 will run at second 0,2,4 and job 1 will run at second 1, 3, 5, you got the idea

app/Console/Commands/DependsOnOpenseaRateLimit.php
```diff
@@ -0,0 +1,104 @@
...
+        }
+
+        $delayInSeconds = $this->getDelayInSeconds($job, $index);
+
+        $dispatchAt = Carbon::now()->addSeconds($delayInSeconds);
+
+        // Note: I cant use delay directly on the job because it throws
+        // the "too many attempts" error after some time so im delaying
+        // with a queued closure instead
+        dispatch($callback)->delay($dispatchAt);
```
oXtxNt9U, line 102: If I understand correctly, there's one potential issue.

Consider this:

FetchCollectionFloorPrice scheduled for 00:10:00
FetchCollectionOpenseaSlug scheduled for 00:05:00

```
$delayInSeconds = $this->getDelayInSeconds($job, $index);
$dispatchAt = Carbon::now()->addSeconds($delayInSeconds);
```

Since this uses `Carbon::now()` and the 2 jobs are scheduled only 5 minutes apart, they still might potentially dispatch overlapping jobs if there are enough collections to result in a spread over 5+ minutes and if there's any latency from calls to e.g. postgres and/or redis that would effectively cancel out the artifical rate limit factor.

In this case the solution is to simply round `Carbon::now()` first based on the job to the nearest "interval":

For example: 

```
FetchCollectionFloorPrice:

Carbon::now() | nearest
00:10:01        00:10:00

FetchCollectionOpenseaSlug:
Carbon::now() | Nearest
00:15:01        00:15:00
```

Now when adding `$delayInSeconds` it should always result in non-overlapping jobs regardless of any latency.

alfonsobries: @oXtxNt9U good catch 🧠

alfonsobries:
updated @oXtxNt9U https://github.com/ArdentHQ/dashbrd/pull/169/commits/958fc3391b041275fe6f24734904e57ec7a26192

app/Console/Commands/DependsOnOpenseaRateLimit.php
```diff
@@ -36,15 +36,23 @@ trait DependsOnOpenseaRateLimit
...
 
         $delayInSeconds = (int) floor($index / $maxRequests) * $perSeconds * $this->getRateLimitFactor();
 
-        return $delayInSeconds + $this->getDelayTreshold($job);
+        return Carbon::now()
+            // Round to the next minute so the delay is always calculated from
+            // the beginning of the minute which prevents overlapping considering
+            // jobs may be dispatched at different times
+            // e.g. 12:00:34 -> 12:01:00, 12:00:59 -> 12:01:00
+            ->endOfMinute()->addSecond()
```
oXtxNt9U, line 52: I think this should be `Carbon::now()->addMinute()->startOfMinute()`

Since there's a subtle difference:

```php

> Carbon\Carbon::now()
= Carbon\Carbon @1697074870 {#8779
    date: 2023-10-12 01:41:10.295788 UTC (+00:00),
  }

> Carbon\Carbon::now()->endOfMinute()
= Carbon\Carbon @1697074919 {#8781
    date: 2023-10-12 01:41:59.999999 UTC (+00:00),
  }

> Carbon\Carbon::now()->endOfMinute()->addSecond()
= Carbon\Carbon @1697074920 {#8777
    date: 2023-10-12 01:42:00.999999 UTC (+00:00),
  }

> Carbon\Carbon::now()->addMinute()->startOfMinute()
= Carbon\Carbon @1697074920 {#8770
    date: 2023-10-12 01:42:00.0 UTC (+00:00),
  }

```

oXtxNt9U (changes requested):

alfonsobries:
@oXtxNt9U updated ✅

oXtxNt9U (approved):

ItsANameToo (approved):

PR #176, opened 2023-10-02
refactor: soft delete gallery nfts

alfonsobries:
## Summary

There is not really easy to test so a careful analysis to the querys is goint to be necessary

Whats expected is the following:

1. When a gallery nft is unassigned from a wallet (no longer belong to wallet) is soft deleted from the gallery and should not appear in the gallery view page
2. if a nft is assigned again it appears in the gallery again
3. if all nfts from the gallery are unassigned the gallery is also soft delted
4. if the nft belongs to the gallery again the gallery is recovered with that nft
5. if the nft changes to another owner is completely removed from the gallery
6. if all nfts are completely removed from a gallery the gallery is also completely removed

https://app.clickup.com/t/8678u37n1

## Checklist

-   [ ] Ready to be merged

### DIFF
--- a/app/Http/Controllers/MyGalleryController.php
+++ b/app/Http/Controllers/MyGalleryController.php
@@ -109,7 +109,7 @@ class MyGalleryController extends Controller
     {
         $this->authorize('delete', $gallery);
 
-        $gallery->delete();
+        $gallery->forceDelete();
 
         return redirect()
             ->to(route('my-galleries'))
--- a/queries/migrations.create_prune_galleries_trigger.sql
+++ b/queries/migrations.create_prune_galleries_trigger.sql
@@ -43,7 +43,6 @@ AND
         WHERE galleries.id = soft_deleted_nfts.gallery_id
     );
 
-
 -- Lastly, delete all empty galleries of old user
 DELETE FROM galleries
 WHERE
--- a/queries/migrations.create_prune_galleries_trigger.sql
+++ b/queries/migrations.create_prune_galleries_trigger.sql
@@ -11,17 +11,29 @@ SELECT g.user_id INTO _user_id FROM galleries g
                                         JOIN wallets w ON g.user_id = w.user_id AND w.id = OLD.wallet_id
     LIMIT 1;
 
--- Cleanup pivot table
-WITH removed_nfts AS (
+-- Mark all gallery nfts as soft deleted if no longer owned by any user
+WITH unassinged_nfts AS (
     SELECT gallery_id, nft_id FROM galleries
                                        JOIN nft_gallery ng on galleries.id = ng.gallery_id
-                                       JOIN nfts n on ng.nft_id = n.id AND (n.wallet_id IS NULL OR n.wallet_id != OLD.wallet_id OR n.deleted_at IS NOT NULL)
+                                       JOIN nfts n on ng.nft_id = n.id AND n.wallet_id IS NULL
     WHERE user_id = _user_id
 )
 UPDATE nft_gallery
     SET deleted_at = NOW()
-FROM removed_nfts rm
+FROM unassinged_nfts un
+WHERE nft_gallery.gallery_id = un.gallery_id AND nft_gallery.nft_id = un.nft_id;
+
+-- remove gallery nfts relationship if nft belongs to another user or is deleted
+WITH removed_nfts AS (
+    SELECT gallery_id, nft_id FROM galleries
+                                       JOIN nft_gallery ng on galleries.id = ng.gallery_id
+                                       JOIN nfts n on ng.nft_id = n.id AND (n.wallet_id != OLD.wallet_id OR n.deleted_at IS NOT NULL)
+    WHERE user_id = _user_id
+)
+DELETE FROM nft_gallery
+    USING removed_nfts rm
 WHERE nft_gallery.gallery_id = rm.gallery_id AND nft_gallery.nft_id = rm.nft_id;
+
 
 -- Mark all galleries with only soft deleted nfts as soft deleted
 WITH soft_deleted_nfts AS (
--- a/queries/migrations.create_prune_galleries_trigger.sql
+++ b/queries/migrations.create_prune_galleries_trigger.sql
@@ -4,7 +4,7 @@
 
 CREATE OR REPLACE FUNCTION prune_empty_galleries() RETURNS TRIGGER AS $_$
 DECLARE
-_user_id BIGINT;
+    _user_id BIGINT;
 BEGIN
     -- Fetch user id of old wallet
 SELECT g.user_id INTO _user_id FROM galleries g
@@ -33,6 +33,8 @@ WITH soft_deleted_nfts AS (
     SELECT gallery_id
     FROM nft_gallery
     WHERE deleted_at IS NOT NULL
+    AND nft_gallery.nft_id = NEW.id
+    AND NEW.wallet_id IS NULL
     GROUP BY gallery_id
     HAVING COUNT(*) = COUNT(deleted_at)
 )
--- a/queries/migrations.create_prune_galleries_trigger.sql
+++ b/queries/migrations.create_prune_galleries_trigger.sql
@@ -28,27 +28,19 @@ DELETE FROM nft_gallery
 WHERE nft_gallery.gallery_id = rm.gallery_id AND nft_gallery.nft_id = rm.nft_id;
 
 
--- Mark all galleries with only soft deleted nfts as soft deleted
-WITH soft_deleted_nfts AS (
-    SELECT gallery_id
-    FROM nft_gallery
-    WHERE deleted_at IS NOT NULL
-    AND nft_gallery.nft_id = NEW.id
-    AND NEW.wallet_id IS NULL
-    GROUP BY gallery_id
-    HAVING COUNT(*) = COUNT(deleted_at)
-)
-
+-- Mark all galleries with all nfts relationship soft deleted as soft deleted
 UPDATE galleries
-    SET deleted_at = NOW()
-WHERE
-    user_id = _user_id
-AND
-    EXISTS (
-        SELECT 1
-        FROM soft_deleted_nfts
-        WHERE galleries.id = soft_deleted_nfts.gallery_id
-    );
+SET deleted_at = NOW()
+WHERE id IN (
+    SELECT DISTINCT ng.gallery_id
+    FROM nft_gallery ng
+    WHERE ng.nft_id = NEW.id OR ng.nft_id = OLD.id
+)
+AND NOT EXISTS (
+    SELECT 1
+    FROM nft_gallery ng
+    WHERE ng.gallery_id = galleries.id AND ng.deleted_at IS NULL
+);
 
 -- Lastly, delete all empty galleries of old user
 DELETE FROM galleries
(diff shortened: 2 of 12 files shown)


app/Http/Controllers/MyGalleryController.php
```diff
@@ -109,7 +109,7 @@ public function destroy(Gallery $gallery): RedirectResponse
     {
         $this->authorize('delete', $gallery);
 
-        $gallery->delete();
+        $gallery->forceDelete();
```
alfonsobries, line 112: since gallery  now has a deleted_at column it need to be forced deleted when deleted by the user

queries/migrations.create_prune_galleries_trigger.sql
```diff
@@ -11,17 +11,50 @@ SELECT g.user_id INTO _user_id FROM galleries g
...
+SET deleted_at = NOW()
+WHERE
+    user_id = _user_id
+AND
+    EXISTS (
+        SELECT 1
+        FROM soft_deleted_nfts
+        WHERE galleries.id = soft_deleted_nfts.gallery_id
+    );
+
```
oXtxNt9U, line 57: The changes here affect an existing trigger. We need a migration for this or the existing database won't see the changes.

Either we do it manually, or put something like this into a separate migration that runs after the original one:

```sql
BEGIN;

DROP TRIGGER IF EXISTS prune_empty_galleries_on_nft_change on nfts;
CREATE TRIGGER prune_empty_galleries_on_nft_change
    AFTER UPDATE ON nfts
    FOR EACH ROW WHEN ( (OLD.wallet_id IS NOT NULL AND NEW.wallet_id IS NULL) OR (OLD.wallet_id != NEW.wallet_id) OR (OLD.deleted_at IS NULL AND NEW.deleted_at IS NOT NULL))
    EXECUTE PROCEDURE prune_empty_galleries();

COMMIT;
```

oXtxNt9U (changes requested):
Will give it a proper review later

ItsANameToo:
To test this (generally):
- have a wallet with some NFTs
- adjust the `FetchWalletNfts` job to return an empty response (or to return a subset based on what you want to test)
- run it manually so it updates the wallet to clear them

Make sure to adjust the `cleanupNftsAndGalleries` method to not ignore your local testing address if you're using that, otherwise it won't affect it

queries/migrations.create_prune_galleries_trigger.sql
alfonsobries: @oXtxNt9U good catch, [fixed here](https://github.com/ArdentHQ/dashbrd/pull/176/commits/6fc0c9ab5c7c4d260b1a8d465a28dfec2bcf568c)

queries/migrations.create_prune_galleries_trigger.sql
```diff
@@ -11,17 +11,50 @@ SELECT g.user_id INTO _user_id FROM galleries g
...
+    SELECT gallery_id, nft_id FROM galleries
+                                       JOIN nft_gallery ng on galleries.id = ng.gallery_id
+                                       JOIN nfts n on ng.nft_id = n.id AND n.wallet_id IS NULL
+    WHERE user_id = _user_id
+)
+UPDATE nft_gallery
+    SET deleted_at = NOW()
+FROM unassinged_nfts un
+WHERE nft_gallery.gallery_id = un.gallery_id AND nft_gallery.nft_id = un.nft_id;
+
```
oXtxNt9U, line 25: Since this trigger is executed per nft row, we do not need to select all nfts each time. It takes 1-2 seconds for me on a `composer db:live` table when removing the wallet for all nfts at once.

We should be able to simplify this to:

```sql
UPDATE nft_gallery
SET deleted_at = NOW()
WHERE nft_gallery.nft_id = NEW.id AND NEW.wallet_id IS NULL;
```

Didn't notice any functional difference, except that it makes it instant.

queries/migrations.create_prune_galleries_trigger.sql
```diff
@@ -11,17 +11,50 @@ SELECT g.user_id INTO _user_id FROM galleries g
...
+                                       JOIN nfts n on ng.nft_id = n.id AND (n.wallet_id != OLD.wallet_id OR n.deleted_at IS NOT NULL)
     WHERE user_id = _user_id
 )
 DELETE FROM nft_gallery
     USING removed_nfts rm
 WHERE nft_gallery.gallery_id = rm.gallery_id AND nft_gallery.nft_id = rm.nft_id;
 
+
+-- Mark all galleries with only soft deleted nfts as soft deleted
+WITH soft_deleted_nfts AS (
```
oXtxNt9U, line 39: Since we know which galleries were affected inside this trigger, we don't need to check all galleries on each invocation.

Instead  we can store the distinct affected gallery ids (see my other comment) via `RETURNING` in a variable to only check those here with something like `WHERE deleted_at IS NOT NULL AND id = ANY (_updated_galleries)`.

oXtxNt9U (changes requested):
I tested manually and it generally seems to work fine. 

Left two comments to improve the queries a bit.

queries/migrations.create_prune_galleries_trigger.sql
alfonsobries: @oXtxNt9U updated ✅

queries/migrations.create_prune_galleries_trigger.sql
alfonsobries: Tried to apply that suggestion but had some sql errrors that was unable to fix (code and error bwlow) apparently you can't directly collect multiple returned values into an array using RETURNING

What I did instead was to add an extra condition in the query to filter the galleries, [see here](https://github.com/ArdentHQ/dashbrd/pull/176/commits/9ec08a25d46dd78ed45d52476d822134cd5cf92d)

What do you think? 

error:

```
: Invalid text representation: 7 ERROR:  malformed array literal: "9"
DETAIL:  Array value must start with "{" or dimension information.
CONTEXT:  PL/pgSQL function prune_empty_galleries() line 12 at SQL statement (Connection: pgsql, SQL: update "nfts" set "wallet_id" = ?, "updated_at" = 2023-10-04 15:15:40 where "wallet_id" = 21 and exists (select * from "collections" where "nfts"."collection_id" = "collections"."id" and "network_id" = 53 and "address" = 0x0053399124f0cbb46d2cbacd8a89cf0599974963 and "collections"."deleted_at" is null) and "updated_at" < 2023-10-04 15:15:40 and "nfts"."deleted_at" is null)
````

code

```sql
CREATE OR REPLACE FUNCTION prune_empty_galleries() RETURNS TRIGGER AS $_$
DECLARE
    _user_id BIGINT;
    _updated_galleries BIGINT[];
BEGIN
    -- Fetch user id of old wallet
    SELECT g.user_id INTO _user_id FROM galleries g
                                        JOIN wallets w ON g.user_id = w.user_id AND w.id = OLD.wallet_id
    LIMIT 1;

    -- Mark all gallery nfts as soft deleted if no longer owned by any user
    UPDATE nft_gallery
    SET deleted_at = NOW()
    WHERE nft_gallery.nft_id = NEW.id AND NEW.wallet_id IS NULL
    RETURNING nft_gallery.gallery_id INTO _updated_galleries;

    -- remove gallery nfts relationship if nft belongs to another user or is deleted
    -- ... (rest of your function here)

    -- Mark all galleries with only soft deleted nfts as soft deleted
    WITH soft_deleted_nfts AS (
        SELECT gallery_id
        FROM nft_gallery
        WHERE deleted_at IS NOT NULL
        AND gallery_id = ANY (_updated_galleries)
        GROUP BY gallery_id
        HAVING COUNT(*) = COUNT(deleted_at)
    )
    UPDATE galleries
        SET deleted_at = NOW()
    WHERE
        user_id = _user_id
    AND
        EXISTS (
            SELECT 1
            FROM soft_deleted_nfts
            WHERE galleries.id = soft_deleted_nfts.gallery_id
        );

    -- ... (rest of your function here)

    RETURN NEW;
END;
$_$ LANGUAGE plpgsql;
```

alfonsobries:
@oXtxNt9U marking as ready but need to be rechecked,

queries/migrations.create_prune_galleries_trigger.sql
oXtxNt9U: @alfonsobries Yeah, I think how you did it is also fine. We can keep it as is then. :+1: 

---

Just for completeness, this is how to use it with an array:

```patch
diff --git a/queries/migrations.create_prune_galleries_trigger.sql b/queries/migrations.create_prune_galleries_trigger.sql
index a87b8d48..1285053d 100644
--- a/queries/migrations.create_prune_galleries_trigger.sql
+++ b/queries/migrations.create_prune_galleries_trigger.sql
@@ -5,6 +5,7 @@
 CREATE OR REPLACE FUNCTION prune_empty_galleries() RETURNS TRIGGER AS $_$
 DECLARE
     _user_id BIGINT;
+    _updated_galleries BIGINT[];    
 BEGIN
     -- Fetch user id of old wallet
 SELECT g.user_id INTO _user_id FROM galleries g
@@ -12,9 +13,15 @@ SELECT g.user_id INTO _user_id FROM galleries g
     LIMIT 1;
 
 -- Mark all gallery nfts as soft deleted if no longer owned by any user
-UPDATE nft_gallery
-SET deleted_at = NOW()
-WHERE nft_gallery.nft_id = NEW.id AND NEW.wallet_id IS NULL;
+WITH updated_galleries AS (
+    UPDATE nft_gallery
+    SET deleted_at = NOW()
+    WHERE nft_gallery.nft_id = NEW.id AND NEW.wallet_id IS NULL
+    RETURNING nft_gallery.gallery_id
+)
+SELECT array_agg(gallery_id)
+FROM updated_galleries 
+INTO _updated_galleries;
 
 -- remove gallery nfts relationship if nft belongs to another user or is deleted
 WITH removed_nfts AS (
@@ -27,14 +34,12 @@ DELETE FROM nft_gallery
     USING removed_nfts rm
 WHERE nft_gallery.gallery_id = rm.gallery_id AND nft_gallery.nft_id = rm.nft_id;
 
-
 -- Mark all galleries with only soft deleted nfts as soft deleted
 WITH soft_deleted_nfts AS (
     SELECT gallery_id
     FROM nft_gallery
     WHERE deleted_at IS NOT NULL
-    AND nft_gallery.nft_id = NEW.id
-    AND NEW.wallet_id IS NULL
+    AND nft_gallery.gallery_id = ANY(_updated_galleries)
     GROUP BY gallery_id
     HAVING COUNT(*) = COUNT(deleted_at)
 )

```

oXtxNt9U (approved):
LGTM.

Someone else should still give it a lookt in case something was missed.

ItsANameToo:
concrete testing steps:
- in `FetchWalletNfts` provide an empty array to store, e.g. `$nftHandler->store(collect([]), true);` and set `nextToken` to `null` if you are working with an address with more NFTs that could cause pagination
- for partial responses, change it to `$nftHandler->store($result->nfts->take(4), true);` where 4 is an arbitrary amount of NFTs you want it to contain
- in all case, mind `cleanupNftsAndGalleries` to not ignore your local address if you're using that

samharperpittam (changes requested):
@alfonsobries 

I have a Gallery with 2 NFT's (my wallet only owns 2 NFT's)
image

I modified  `$nftHandler->store($result->nfts, true);` to `$nftHandler->store($result->nfts->take(1), true);` and then fetched nfts for my wallet. The intention of this change was to see if 1 NFT would be removed from my Gallery.

However, once I had fetched nfts for my wallet, it appears the whole Gallery was removed from Dashbrd. 

image

But dashbrd is still reporting that my wallet owns 1 nft
image

In this scenario I would have expected the Gallery to remain but just with 1 NFT.

If I revert changes and then fetch nfts for my wallet, the Gallery is restored. But it feels like with a partial deletion, we're removing the whole gallery rather than just the nfts that no longer exist for a wallet?

ItsANameToo:
I can replicate the above issue. Created a gallery with 2 NFTs, then made it return 1 on the response. One NFT is still assigned in the database at the moment, but the gallery was softdeleted.

ItsANameToo (approved):

PR #182, opened 2023-10-03
feat: user crud and permissions

alfonsobries:
## Summary

- Crud for users
- For roles im only adding the permission to assign permissions to existing roles, cannot add or delete roles which is what I think that makes more sense

https://app.clickup.com/t/8678v094q

## Checklist

-   [ ] Ready to be merged


app/Filament/Resources/UserResource/Pages/Traits/HandleRole.php
```diff
@@ -0,0 +1,32 @@
...
+use Illuminate\Database\Eloquent\Model;
+
+trait HandleRole
+{
+    private function handleRole(Model $model, string $role): void
+    {
+        /** @var User */
+        $user = auth()->user();
+
+        $validRoles = [Role::Editor->value];
```
ItsANameToo, line 18: we need a way to add roles as current demo/prod have been seeded prior to these changes and only include superadmin/admin

app/Filament/Resources/UserResource.php
```diff
@@ -4,47 +4,132 @@
...
+                    ->required(),
+
+                SpatieMediaLibraryFileUpload::make('avatar')
+                    ->collection('avatar')
+                    ->columnSpan('full')
+                    ->image()
+                    ->imageEditor()
+                    ->imageCropAspectRatio('1:1'),
+
+                CheckboxList::make('permissions')
```
ItsANameToo, line 82: can you add a subtitle/helper text to indicate that these permissions will be on an individual basis and in addition to the ones from the assigned role, e.g:
"Selected permissions here will be applied in addition to the assigned user role"

database/migrations/2023_10_05_153934_sync_permissions_and_roles.php
```diff
@@ -0,0 +1,55 @@
...
+        $roles = config('permission.roles');
+
+        $roles = collect($roles)->map(static fn ($permissions, $role) => [
+            'name' => $role,
+            'guard_name' => 'admin',
+            'created_at' => now(),
+            'updated_at' => now(),
+        ])->all();
+
+        Role::upsert($roles, ['name', 'guard_name'], []);
```
ItsANameToo, line 44: can we assign all the permissions to the roles in the migration as well? then we don't have to go in and assign them manually when reseeding locally + will make it easier in production as well if everything is setup correctly immediately

ItsANameToo (changes requested):

database/migrations/2023_10_05_153934_sync_permissions_and_roles.php
alfonsobries: @ItsANameToo the problem with doing that is that is going to override any previous permissions assigned by the user (maybe thats fine)

locally the seeders still run the permissions seeder that adds the default permissions so should not be a problem locally so depends on whether we want or not to override any previously assigned permission in the migration, in my mind feels cleaner to keep that on the seeder but also not a problem since there permissions are going to be defined for the first time, what do you think?

ItsANameToo: right now there are no customisations made, so we can safely adjust them still. I'd have the migration overwrite it for now as it won't affect us on demo/prod and we can re-assess this for subsequent releases as needed

ItsANameToo: and it makes it less error-prone than manually assigning the permissions to the roles

alfonsobries:
@ItsANameToo now the default permissions are set on the migrations and removed the permission seeder

ItsANameToo (approved):

PR #183, opened 2023-10-05
refactor: add the button to refresh the collections

crnkovic:
## Summary

https://app.clickup.com/t/862kebafh

## Checklist

-   [ ] Ready to be merged


app/Http/Controllers/RefreshedCollectionController.php
```diff
@@ -0,0 +1,56 @@
...
+
+        $wallet->update([
+            'is_refreshing_collections' => true,
+        ]);
+
+        Bus::batch($collections->flatMap(fn ($collection) => [
+            new FetchCollectionFloorPrice($collection->network->chain_id, $collection->address),
+            new FetchCollectionTraits($collection),
+            new FetchCollectionOwners($collection),
+            new FetchCollectionVolume($collection),
```
oXtxNt9U, line 41: I think it makes sense to somehow impose a global rate limit here, not just per wallet because there can be overlap in the collections between wallets.

While we already should have limits on each individual collection job, having a dedicated job makes it more flexible and extensible later on. It also improves observability, since we can correlate jobs more easily.

So maybe we should just wrap the individual dispatches in a dedicated job like `RefreshCollection` and this job then can be rate limited similar to the `RefreshNftMetadata` one. Just pick any sensible default.

You can check how it's done here:

https://github.com/ArdentHQ/dashbrd/blob/ee4f2ac5e16b60d71345c79aaca76f5dc002b1d2/app/Http/Controllers/RefreshedNftController.php#L22

https://github.com/ArdentHQ/dashbrd/blob/ee4f2ac5e16b60d71345c79aaca76f5dc002b1d2/app/Providers/AppServiceProvider.php#L67

The cooldown on the wallet, etc. can stay as is I think.

oXtxNt9U (changes requested):

crnkovic:
One thing that was discovered by @oXtxNt9U is if the wallet has a lot of collections, updating all of them potentially takes a long time, with all the rate limits... and every time the user refreshes the page, we'll show the toast that we're refreshing their collections, until refreshing ends. In my mind, this creates worse UX as users can potentially think something is buggy if we show the toast for an hour or even longer.

Personally, I'd just ditch the notification. We already guard against spamming the refresh button in the backend, so we can potentially even remove the tooltip and let them click the refresh button all they want.

cc @boldninja @ItsANameToo.

ItsANameToo:
in that case the toast should disappear after a few seconds so it's only used to inform the user something is happening without it reflecting the time that's needed for it 🤔 that way you click -> get a toast to indicate it will be refreshed and user will notice when data has been updated later on.

Going off goga's video, I'd try to avoid refreshing the page when that button is clicked as that can give the false impression that refreshing immediately pulls in the new data on the page

app/Http/Controllers/RefreshedCollectionController.php
```diff
@@ -0,0 +1,26 @@
...
+
+class RefreshedCollectionController extends Controller
+{
+    public function store(Request $request): RedirectResponse
+    {
+        abort_unless($request->wallet()->canRefreshCollections(), 403);
+
+        RefreshWalletCollections::dispatch($request->wallet())->onQueue(Queues::WALLETS);
+
+        return back()->toast(
```
oXtxNt9U, line 20: @crnkovic Not sure exactly about this, but I'm still getting redirected to the landing page whenever I hit the refresh button. Does it not happen for you?

oXtxNt9U (changes requested):

app/Http/Controllers/RefreshedCollectionController.php
crnkovic: This is how it looks for me.

https://github.com/ArdentHQ/dashbrd/assets/6536260/e280a21f-a27f-4f1a-8f08-2099d6413d2c

resources/js/Pages/Collections/Components/CollectionsHeading/RefreshButton.tsx
```diff
@@ -0,0 +1,74 @@
...
+    const [loading, setLoading] = useState(false);
+    const [disabled, setDisabled] = useState(false);
+    const { t } = useTranslation();
+
+    const { signedAction } = useAuthorizedAction();
+    const { showToast } = useToasts();
+
+    const refresh = (): void => {
+        void signedAction(async ({ authenticated }) => {
+            if (authenticated) {
```
alfonsobries, line 19: The `authenticated` parameter (or, in some cases, you can also pass the `signed` parameter) represents the state before signing in or logging in. One use case for this parameter is when a user attempts to like a gallery while not logged in. In such cases, we don't know whether the user had previously liked the gallery, so this parameter is used to force a like action instead of toggling the value.

In this specific case, I believe we don't need to check for the authenticated parameter (you can't even access this section if you are not authenticated). Therefore, you can remove that condition. In other words, if the method runs, it implies the user is authenticated AND signed.

Side note: Perhaps we should consider renaming these parameters to usedToBeAuthenticated and usedToBeSigned to avoid confusion. What do you think?

alfonsobries (changes requested):

oXtxNt9U (approved):

database/migrations/2023_10_04_142614_add_refreshing_collections_columns_to_wallets_table.php
```diff
@@ -0,0 +1,21 @@
...
+return new class extends Migration
+{
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        Schema::table('wallets', function (Blueprint $table) {
+            $table->boolean('is_refreshing_collections')->default(false)->after('extra_attributes');
+            $table->timestamp('refreshed_collections_at')->nullable()->after('is_refreshing_collections');
```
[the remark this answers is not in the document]
crnkovic: honestly I think `extra_attributes` column just creates a headache for us (getting data from json, updating data in json). adding new columns to the table doesn't have much downsides

ItsANameToo (approved):

PR #188, opened 2023-10-06
[wip] gallery dynamic meta image

alfonsobries:
## Checklist

-   [ ] Ready to be merged


shahin-hq:
@alfonsobries Is it something we want to expose for public use? If I understood, the goal is to generate meta images for galleries, can we schedule a job on the backend (when gallery updated schedule a job) instead of exposing to public?

app/Console/Commands/PruneMetaImages.php
```diff
@@ -0,0 +1,54 @@
...
+
+        $directory = storage_path(sprintf('meta/galleries/'));
+
+        if (is_dir($directory)) {
+            // Open directory
+            if ($handle = opendir($directory)) {
+                // Loop through each file in the directory
+                while (false !== ($file = readdir($handle))) {
+                    // Check if file is a PNG and not in validImages
+                    if (pathinfo($file, PATHINFO_EXTENSION) === 'png' && ! in_array($file, $validImages)) {
```
oXtxNt9U, line 41: Just to future proof it might be better to use `isset` O(1) because `in_array` potentially has to walk the entire array for each file in the directory O(m*n).

https://stackoverflow.com/a/13483548

app/Http/Controllers/MetaImageController.php
```diff
@@ -0,0 +1,100 @@
...
+        return ! file_exists($imagePath);
+    }
+
+    private function getImagePath(Gallery $gallery): string
+    {
+        return storage_path(sprintf('meta/galleries/%s.png', $this->getImageName($gallery)));
+    }
+
+    /**
+     * IMPORTANT: Ensure this method is in sync with the query in `queries/gallery.get_meta_images.sql`
```
oXtxNt9U, line 47: Confirmed that both implementations give the same result.

oXtxNt9U (approved):
@alfonsobries Small nitpick, but overall looks good to me :+1: 

Will already approve, though it should get another review or two.

app/Console/Commands/PruneMetaImages.php
```diff
@@ -0,0 +1,56 @@
...
+     * @var string
+     */
+    protected $signature = 'prune-meta-images';
+
+    /**
+     * The console command description.
+     *
+     * @var string
+     */
+    protected $description = 'Removes unusued meta images';
```
crnkovic, line 24:
```suggestion
    protected $description = 'Removes unused meta images';
```

app/Console/Commands/PruneMetaImages.php
```diff
@@ -0,0 +1,56 @@
...
+
+    /**
+     * Execute the console command.
+     */
+    public function handle(): int
```
crnkovic, line 29: not the prettiest method, but we can clear it up with time

crnkovic (approved):

crnkovic:
One thing: this PR requires you to install Puppeteer. Should we add that library to `package.json` so it doesn't rely on Puppeteer being installed globally on the host machine?

ItsANameToo:
> One thing: this PR requires you to install Puppeteer. Should we add that library to `package.json` so it doesn't rely on Puppeteer being installed globally on the host machine? cc @ItsANameToo

if there is no downside of having it installed by a different user compared to a global installation (permission wise) then package.json would be easier yeah

alfonsobries:
@ItsANameToo @crnkovic added puppeter as dependency, works fine locally without any extra adjustment (first i removed puppeter from my global dependencies, tested (didnt work), then added the local dependency and worked again)

crnkovic (approved):

PR #200, opened 2023-10-09
refactor: session timeout handling

alfonsobries:
## Summary

Im not able to replicate the issue as the video on the ticket but I think that is because we currently handle the "signed actions" differently, still I can see an issue after session expired: if you try to do something that requires a signature the request returns a 403 error that could be invisible to the user, for example if you sign your wallet, wait until session expires and you try to like a gallery you will see an error in the console but no indication that something wrong happen in the frontend (other that the count not be updated)

Tried different alternatives but think the best one is to check the signed/authenticated status with the backed, which means an extra request every time the user tries to do something that requires a signature, the issue is that it represents a quick delay but personally think is not something that really affects the user experience but that subject to debate, 

Closes https://app.clickup.com/t/862kf99rz

## Checklist

-   [ ] Ready to be merged


ItsANameToo:
some test is hanging and runs into the CI limit @alfonsobries

ItsANameToo (changes requested):
question, would it be possible to handle the response from the backend instead where a user has no active session anymore and handle it accordingly? Then we can keep the signed status as is as that's what will work for majority of the time, but have an additional check in case the response from the backend fails due to the session being expired?

That would allow us to have a scenario where you like a gallery and get prompted to sign again with a "session expired" message so the user can sign it and continue as opposed to having to request this each time while majority of users will have a valid session

alfonsobries:
@ItsANameToo I tried that at first but it was a complex problem, I dont remember all the exact scenarios that I needed to take care but remember that it meant some tricks and stuff, but will spend some time again to try to find a solution and let you know

alfonsobries:
closing in favour of https://github.com/ArdentHQ/dashbrd/pull/231 cc @ItsANameToo

PR #203, opened 2023-10-10
fix: make collections grid view respect list view sorting

patricio0312rev:
# [[collections] grid view does not respect list view sorting order](https://app.clickup.com/t/862kfhg2v)

## Summary

- When using sorting options, the grid view now respects the order provided by the BE.
- When moving back to list view, the list view also respects the order provided in the URL.

- This happened because the grid view was resorting all the results provided by the BE.
- In the case of the list view, `manualSortBy` prop has been set to `true`, since we are pre-sorting the data in the BE, just as the [`react-table` documentation](https://tanstack.com/table/v8/docs/api/features/sorting#manualsorting) mentions.

## Steps to reproduce

1. Run the app in `local` or `testing_e2e`.
2. Go to `/collections` page.
3. Click on `Sort` button and select any option.
4. Toggle to grid view or list view and see the magic ✨ 

https://github.com/ArdentHQ/dashbrd/assets/55117912/62ad77df-6927-4580-b508-cbfe7138c454

## Checklist

-   [ ] Ready to be merged


resources/js/Components/Collections/CollectionsGrid/CollectionsGrid.test.tsx
```diff
@@ -90,19 +90,19 @@ describe("CollectionsGrid", () => {
...
-            "0.2 ETH",
+            "700 USDC",
             "300 USDC",
-            "200 USDC",
-            "1 ETH",
-            "1 ETH",
-            "100 USDC",
             "100 USDC",
-            "0.05 ETH",
             "50 USDC",
```
ItsANameToo, line 106: I'm confused about this "sorted" order as it doesn't seem sorted by value 🤔 care to explain what is expected here in terms of sorting? To me it seems to be identical to the original array values and not sorted:

```
const ethereumValues = [0.1, 0.2, 0.4, 0.7, 0.3, 0.1, 0.05, null, null];
const usdcValues = [100, 200, 400, 700, 300, 100, 50, null, null];
```

ItsANameToo: but if it's sorted by value and value is floor price * currency * nft amount, where nfts are always 1, shouldn't the same eth values be together? now it's:
```
"0.1 ETH", // -> 0.1
"0.2 ETH",
"0.4 ETH",
"0.7 ETH",
"0.3 ETH",
"0.1 ETH", // -> also 0.1 ?
 "0.05 ETH",
 ```
 
 which makes it hard to read what this test is really doing. Can you add a comment with the values behind it so it's easier to see the sorting order that it tests against?

ItsANameToo (approved):

oXtxNt9U (approved):

crnkovic (approved):

ItsANameToo:
cherry picked to `rc-0.8.0`

PR #205, opened 2023-10-11
feat: articles page filters

shahin-hq:
## Checklist

-   [ ] Ready to be merged


ItsANameToo:
some conflicts @shahin-hq

app/Data/Articles/FeaturedCollectionData.php
```diff
@@ -16,6 +16,7 @@ class FeaturedCollectionData extends Data
 {
     public function __construct(
         public string $name,
+        public string $slug,
```
ItsANameToo, line 19: where is this used for featured collections? they don't link to anything 🤔

app/Models/Article.php
```diff
@@ -108,4 +148,25 @@ public function getSlugOptions(): SlugOptions
...
+    {
+        $now = now();
+        $pastWeek = now()->subDays(7);
+
+        Article::query()
+            ->update([
+                'views_count_7days' => DB::raw(
+                    <<<SQL
+(
+    SELECT COUNT(*) FROM views AS v
```
ItsANameToo, line 162: what about using https://github.com/cyrildewit/eloquent-viewable#get-views-count-of-a-specific-period for this as opposed to a raw query

ItsANameToo (changes requested):
noticed it on here, but when I visit a collection and go to the `articles` tab; the articles don't include the current collection in the featured collections preview. If it's unrelated to these changes then let me know and I'll add a separate card for it. 

Moonbirds (missing moonbirds, shows timeless characters):
image

Timeless characters (missing, shows moonbirds)
image

app/Data/Articles/FeaturedCollectionData.php
shahin-hq: Yes, but we will need that here: 

> On the right we display the collection images of the featured collections in the article. On hover we display the collection name and on click it navigates to the collection on Dashbrd. On the left side we show a basic count of the amount of featured collections in the article. Similar to the gallery component that already exists

image

app/Models/Article.php
shahin-hq: if we take the given approach we will have to pull all models (articles) and then do the update:

1. step:
```
// Example: get views count from 2017 upto 2018
views($post)
    ->period(Period::create('2017', '2018'))
    ->count();
```
2.nd step:
```
$posts = Post::all();

foreach($posts as $post) {
    $post->unique_views_count = views($post)->unique()->count();
}
```

While the crafted query will do it on the fly. I can still implement this solution but it will be more resource intensive than the current one.

shahin-hq:
> noticed it on here, but when I visit a collection and go to the articles tab; the articles don't include the current collection in the featured collections preview. If it's unrelated to these changes then let me know and I'll add a separate card for it.

ah, you need that too? I filtered out it on purpose, removing the filter then.
@ItsANameToo

shahin-hq:
@ItsANameToo should be fine now.

ItsANameToo (approved):

PR #207, opened 2023-10-11
refactor: collection controller optimizations

crnkovic:
## Summary

Potentially fixes https://app.clickup.com/t/85zu2rdqj

The problem could potentially be loading NFT/collection traits on the collections index page, even though we don't use traits on that page. 

If you seed a bunch of traits to your collections and attach them to that collection's NFTs then open the collections page, you'll see that we eager load all traits for collections/NFTs and it takes forever to load. For example, I've added a few thousand traits to a collection some of its NFTs, and it was loading all thousands of trait models in memory.

We don't need traits for the collections page, so we can remove it... so this PR only eager loads data that we actually need, plus removes some of the complex queries we have.

Changes done in this PR:

- We have this scope on the collection model `forCollectionData` that generated this massive query to prepare all data for display on the collections page. However, we can eager load all required relationships and avoid using one gigantic query in favor of a few smaller ones. Relations we care about are `network` and `floorPriceToken`.
- Instead of eager loading NFTs, we also built this gigantic query to load NFTs for "currently visible" collections for the wallet; then grouped them by collection and displayed as "related" NFTs for a collection (`Nft::forCollections()` method). We can use [this](https://github.com/staudenmeir/eloquent-eager-limit) library that gives us the ability to limit eager loaded relations, such as: `->with(['nfts' => fn ($q) => $q->limit(4)])`. This essentially creates the similar query, just easier to maintain, plus we can then just use `collection.nfts` in the UI.
- On the collections page, we only care about the NFT ID and image, so I've added a `SimpleNftData` object that only has that data. This was the culprit; here we loaded all traits and bunch of other relations that are not needed for the collections page.

All in all, removing complex queries is aces in my book 🤣

### Numbers

- NFTs: 2.5M
- Wallets: 1000
- Collections: 500
- Total Collection Traits: 5K
- Total NFT traits: 30K

#### Before (_WITHOUT ANY_ collection traits)

First request:
- Booting: 120ms
- Application: 240ms
- Queries: 23
- Models: 11

Second request:
- Booting: 110ms
- Application: 400ms
- Queries: 28
- Models: 373 (+ every trait the collection/NFT has)

#### After (_WITH BUNCH_ of collection traits)

First request:
- Booting: 100ms
- Application: 220ms
- Queries: 22
- Models: 11

Second request:
- Booting: 110ms
- Application: 240ms
- Queries: 30
- Models: 93

## Checklist

-   [ ] Ready to be merged


oXtxNt9U (approved):
Nice changes :100:

ItsANameToo (approved):

PR #219, opened 2023-10-12
refactor: fetch opensea slugs more often

alfonsobries:
## Summary

Closes https://app.clickup.com/t/8678xa6qg

Note: Im not calling the job immediately or as a result of calling the `FetchWalletNfts` job because the reasons I explain on the ticket but instead im calling the command that triggers the job to retrieve opensea slugs every 5 minutes.

Consider that the `FetchCollectionOpenseaSlug` command only calls the jobs that actually make requests in the case it finds new collections without slug, that means that in most cases is only going to run a basic select query every 5 minutes and will not make any request

Think a delay of 5 minutes for fetching the missing slugs should be good enough but if necessary we can adjust the job to run every minute for example, since is a relatively basic query should not affect performance a lot

## Checklist

-   [ ] Ready to be merged


app/Console/Kernel.php
```diff
@@ -104,14 +104,15 @@ private function scheduleJobsForCollectionsOrGalleries(Schedule $schedule): void
                 ->hourlyAt(5);
 
         $schedule
-            ->command(FetchCollectionOpenseaSlug::class)
+            ->command(FetchWalletNfts::class)
             ->withoutOverlapping()
-            ->hourlyAt(10);
+            ->hourlyAt(10); // offset by 10 mins so it's not run the same time as FetchEnsDetails...
```
oXtxNt9U, line 109: Was there any specific reason to change the `FetchWalletNfts` schedule to `hourlyAt(10)` or would this PR work without the change just as well?

Just asking out of curiosity, fine with keeping the change regardless.

oXtxNt9U (commented):
@alfonsobries one question

oXtxNt9U (approved):

ItsANameToo (approved):

PR #230, opened 2023-10-13
feat: handle errors in Alchemy response for getWalletNfts

patricio0312rev:
# [[nfts] handle alchemy nft token uri error](https://app.clickup.com/t/862kfnnbz)

## Summary

- A log has been added in case we receive an NFT with the error property.
- A new optional argument to make the filter for NFTs with errors in Alchemy's response has been added to prevent errors in other implementations of this function.
- New `hasError` optional boolean property set for `Web3NFTdata`.
- New unit tests for these changes and some other tests have been updated.

## Checklist

-   [ ] Ready to be merged


app/Http/Client/Alchemy/AlchemyPendingRequest.php
```diff
@@ -395,11 +396,20 @@ public function parseNft(array $nft, int $networkId, bool $convertTokenNumber =
 
         $tokenNumber = $convertTokenNumber === true ? CryptoUtils::hexToBigIntStr($nft['id']['tokenId']) : $nft['id']['tokenId'];
 
+        $error = Arr::get($nft, 'error');
+        if (! empty($error)) {
+            Log::info('AlchemyPendingRequest: Filter NFT', [
```
oXtxNt9U, line 401: Can you include the collection address, in case it's difficult to infer it from the name alone (due to ambiguity, etc.)

oXtxNt9U (changes requested):

oXtxNt9U (changes requested):

app/Http/Client/Alchemy/AlchemyPendingRequest.php
```diff
@@ -421,6 +441,8 @@ public function parseNft(array $nft, int $networkId, bool $convertTokenNumber =
             traits: $this->extractTraits($nft),
             mintedBlock: $nft['contractMetadata']['deployedBlockNumber'],
             mintedAt: $mintTimestamp !== null ? Carbon::createFromTimestampMs($mintTimestamp) : null,
+            hasError: ! empty($error),
+            error: $nftError,
```
shahin-hq, line 445: @patricio0312rev it seems `error` and `hasError` has different meaning while they look like related, at the first glance I thought `error` would be string that shows the actual error while `hasError` is a flag to show the `error` is not `null`, however, it is not the case, could you please give a better names to these?

oXtxNt9U: Fair point, maybe we can rename `error` to `info`. So it becomes `NftInfo::MetadataOutdated` and `info: $nftInfo` respectively.

Another option could be `hint`.

(The rename should also apply to the newly added column)

database/migrations/2023_10_17_084040_add_error_column_to_nfts_table.php
```diff
@@ -0,0 +1,28 @@
...
+            $table->string('error')->nullable();
+        });
+    }
+
+    /**
+     * Reverse the migrations.
+     */
+    public function down(): void
+    {
+        Schema::table('nfts', function (Blueprint $table) {
```
oXtxNt9U, line 26: We usually don't support reverse migrations. So this should be removed.

oXtxNt9U (changes requested):

oXtxNt9U (approved):

app/Http/Client/Alchemy/AlchemyPendingRequest.php
```diff
@@ -685,13 +707,13 @@ private function getNftV2ApiUrl(): string
         return 'https://'.self::$apiUrlPlaceholder.'.g.alchemy.com/nft/v2/';
     }
 
-    private function filterNft(mixed $nft): bool
+    private function filterNft(mixed $nft, bool $filterError = true): bool
```
ItsANameToo, line 710: what is `filterError` for? I see it used once here https://github.com/ArdentHQ/dashbrd/pull/230/files#diff-957272cbfdd990765850dc7924f1ac272f6ea71fb3403acbe633fe548369ebf6R213 but that doesn't make it clear to me when to pass that value 🤔

ItsANameToo (approved):

shahin-hq (approved):

PR #234, opened 2023-10-16
fix: reset activity state if collection is invalid

crnkovic:
## Summary

It could happen that if activity is "invalid" (spam or something) it will get stuck in this `is_fetching_activity=true` state. Also I increased job duration to accommodate higher number of collections.

## Checklist

-   [ ] Ready to be merged

### DIFF
--- a/app/Console/Commands/FetchCollectionActivity.php
+++ b/app/Console/Commands/FetchCollectionActivity.php
@@ -45,7 +45,7 @@ class FetchCollectionActivity extends Command
         };
 
         $this->forEachCollection(function ($collection) {
-            Job::dispatch($collection);
+            Job::dispatch($collection, forced: true);
         }, $queryCallback);
 
         return Command::SUCCESS;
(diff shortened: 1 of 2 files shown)


ItsANameToo (approved):

app/Console/Commands/FetchCollectionActivity.php
```diff
@@ -45,7 +45,7 @@ public function handle(): int
         };
 
         $this->forEachCollection(function ($collection) {
-            Job::dispatch($collection);
+            Job::dispatch($collection, forced: true);
```
ItsANameToo, line 48: what does this do?

crnkovic: when stress-testing, i discovered that there is a bug with the job... which is if the job is rate limited on the very first attempt, it will be released back onto the queue, but since the `is_fetching_activity` flag has been set and `forced` is `false`, it will not run, being left in this weird state where job didn't run but `is_fetching_activity` is true.

so i set the forced flag to make sure it runs if rate limited on the very first attempt

shahin-hq (approved):

PR #240, opened 2023-10-17
fix: timestamp issue for report a gallery test

patricio0312rev:
# [[general] gallery flaky test](https://app.clickup.com/t/862kg590t)

## Summary

- The `created_at` timestamp for the `report a gallery` test has been fixed.
- This change will make the date more predictable and less sensitive to delays or other timing issues our testing environment can have.
- After this change, the test suite was run +30 times in a row without any errors, locally.

## Steps to reproduce

1. Open your terminal and locate in the project's folder.
2. Run `php artisan test --filter=GalleryTest` and see the magic ✨ 

## Checklist

-   [ ] Ready to be merged


crnkovic (approved):

tests/App/Models/GalleryTest.php
```diff
@@ -382,6 +382,8 @@
 ]);
 
 it('determines if a gallery was reported recently by the user', function ($hoursAgo, $expected) {
+    $reportTimestamp = now()->subHours($hoursAgo);
```
ItsANameToo, line 385: how does moving it here fix the issue? 🤔 the test itself doesn't reuse the timestamp so I don't follow how moving it up would make it less flaky

ItsANameToo: won't moving it up make it be more affected by the user and gallery creation? start of the method it's 48 hours from now, then user + gallery gets created and it's 48 hours + time it took for that to be done. 

Maybe you can have a look whether `Carbon::setTestNow()` would help with this to set it to a specific value to test against, otherwise go with your suggestion of changing the intervals slightly to have more time to run the test in itself

ItsANameToo: further context, test still fails for me on this branch:
```
Tests\App\Models\GalleryTest > it determines if a gallery was reported recently by the user (48, true) @ repetition 22 of 100
  Failed asserting that false is identical to true.
  ```

PR #249, opened 2023-10-18
feat: SyncActivityPrices command to update missing nft activity sales prices

patricio0312rev:
# [[nft activity] compute and sync total usd/native amount from the token price history](https://app.clickup.com/t/862kg3hj8)

## Summary

- New `activities:sync-prices` command to update all the NFT activities from a Polygon collection in our DB.

## Explanation of the command

This new command runs a transaction query that updates the 'nft_activity' table by calculating and setting the 'total_native' and 'total_usd' columns based on the values stored in the 'extra_attributes' column from the 'token_price_history' table and the `price` column in the `token_price_history` table. The command performs the following actions:

1. For each record in 'nft_activity':
     - Extract the 'totalNative' value from the 'extra_attributes' JSON column and set it as 'total_native'.
     - Calculate 'total_usd' by multiplying 'total_native' with the 'ethereum' token price from 'token_price_history' for the time that activity happened.

2. Update is limited to records in 'nft_activity' associated with collections having a 'network_id' of 1 (Polygon in our app).

3. The command uses a transaction to ensure data consistency. It commits the changes after successful execution.

## Steps to reproduce

1. Run the migrations for the app.
2. Make sure `token_price_history`  table is empty.
3. Index a handful of activities (just for Polygon collections).
4. Index token price history for Matic (you can use `php artisan marketdata:fetch-price-history`).
5. Run our new command: `php artisan activities:sync-prices`.
6. Test that `total_native` and `total_usd` is now filled for all the entries that belong to a Polygon collection.

## Checklist

-   [ ] Ready to be merged


app/Console/Commands/SyncActivityPrices.php
```diff
@@ -0,0 +1,86 @@
...
+                                  token_guid = 'ethereum'
+                                  AND timestamp <= nft_activity.timestamp
+                                ORDER BY timestamp DESC
+                                LIMIT 1
+                              )
+                WHERE
+                  collection_id IN (
+                    SELECT collection_id
+                    FROM collections
+                    WHERE network_id = 1
```
oXtxNt9U, line 68: Instead of hardcoding 1 we can use `Chains::ETH->value` here.

app/Console/Commands/SyncActivityPrices.php
```diff
@@ -0,0 +1,86 @@
...
+     * 2. Update is limited to records in 'nft_activity' associated with collections having a 'network_id' of 1 (Polygon in our app).
+     *
+     * 3. The command uses a SQL transaction to ensure data consistency. It commits the changes after successful execution.
+     *
+     * Caution: Before running this command, make sure to run `tokens:live-dump-price-history`.
+     */
+    public function handle(): int
+    {
+        try {
+            DB::beginTransaction();
```
oXtxNt9U, line 47: Technically, I don't think the transaction is needed since we only have a single statement and postgres wraps statements by default in a transaction (subquery is part of it too).

But probably doesn't hurt to keep it.

oXtxNt9U (changes requested):

oXtxNt9U (dismissed):

app/Console/Commands/SyncActivityPrices.php
```diff
@@ -0,0 +1,92 @@
...
+                                FROM token_price_history
+                                WHERE
+                                  token_guid = '{$ethereumGuid}'
+                                  AND timestamp <= nft_activity.timestamp
+                                ORDER BY timestamp DESC
+                                LIMIT 1
+                              )
+                WHERE
+                  collection_id IN (
+                    SELECT collection_id
```
oXtxNt9U, line 72: I missed this earlier, but I think it should be `id` or to be completely safe `collections.id`, otherwise we will always take the outer `nft_activity.collection_id` which might cause updates of unrelated collections from different networks.

oXtxNt9U (changes requested):

ItsANameToo (changes requested):
needs checking how long this takes to run for x amount of records, as the initial reports would indicate it's too slow to run this on the production database

app/Console/Commands/SyncActivityPrices.php
```diff
@@ -0,0 +1,92 @@
...
+                                SELECT price
+                                FROM token_price_history
+                                WHERE
+                                  token_guid = '{$ethereumGuid}'
+                                  AND timestamp <= nft_activity.timestamp
+                                ORDER BY timestamp DESC
+                                LIMIT 1
+                              )
+                WHERE
+                    nft_activity.collection_id IN (
```
ItsANameToo, line 71: you can make this more specific by also checking the activity `type`, as we only need to have the values populated for `LABEL_SALE` activity

Additionally you can check whether any activity already has `total_native` and `total_usd` set and ignore those, as we don't have to set them again (doubt this will filter out a lot of rows, but may still be worth it)

Best to check against the demo dump to see how many rows it returns with the adjustments compared to before to get an idea of how many it will update compared to before

app/Console/Commands/SyncActivityPrices.php
```diff
@@ -0,0 +1,94 @@
...
+            $updateSql = "
+                UPDATE nft_activity
+                SET
+                  total_native = (extra_attributes->'recipientPaid'->>'totalNative')::numeric,
+                  total_usd = (extra_attributes->'recipientPaid'->>'totalNative')::numeric *
+                              (
+                                SELECT price
+                                FROM token_price_history
+                                WHERE
+                                  token_guid = '{$ethereumGuid}'
```
ItsANameToo, line 65: this is a wrong assumption actually, as the values can be in MATIC as well for Polygon. You'd need to check if they always report in the native currency (meaning you can hardcode this to MATIC instead of ETH) or not (meaning you need to determine it from extra_attributes if possible)

ItsANameToo (changes requested):

app/Console/Commands/SyncActivityPrices.php
oXtxNt9U: Seems to be correct now and match the comment here: https://github.com/ArdentHQ/dashbrd/blob/e4ca3ad620dbe4fc63c3d17af465f4f434d67c1b/app/Http/Client/Mnemonic/MnemonicPendingRequest.php#L432

oXtxNt9U (approved):

ItsANameToo (approved):

PR #252, opened 2023-10-18
feat: article content

alfonsobries:
## Summary

Closes https://app.clickup.com/t/8678v0jvy

Note: the content is hardcode, also im not handling external links or code hightlighting, this pr is mostly about applying the design styles

Pending stuff for other tasks:

- Parsing article markdown
- Handling external links
- Highlighting code?

## Checklist

-   [ ] I checked my UI changes against the design and there are no notable differences
-   [ ] I checked my UI changes for any responsiveness issues
-   [ ] I checked my (code) changes for obvious issues, debug statements and commented code
-   [ ] I opened a corresponding card on Clickup for any remaining TODOs in my code
-   [ ] I added a short description on how to test this PR _(if necessary)_
-   [ ] I added a storybook entry for the component that was added _(if necessary)_
-   [ ] Documentation _(if necessary)_
-   [ ] Tests _(if necessary)_
-   [ ] Ready to be merged

PR #271, opened 2023-10-23
feat: article markdown parsing

alfonsobries:
## Summary

Depends on https://github.com/ArdentHQ/dashbrd/pull/252

Closes https://app.clickup.com/t/865d922px.

Uses [react-markdown](https://www.npmjs.com/package/react-markdown) to parse the markdown.

The plugin works well in the sense that it seems to parse the content of the test articles correctly. However, the content doesn't appear as we might expect it to. This is more like an issue with the articles that were prepared for MSQ rather than a problem with the parser itself. For example, the articles have images, and it appears that an attempt was made to add a "caption" using the blockquote format `> ...`. In MSQ, this is handled differently, so it appears correctly there, but here, since it's the default format, it generates a literal blockquote block.

image

In my opinion, it's fine to leave it as it is because it avoids adding patches or stuff like that, which allows us to switch between parsers if necessary, and in general, it will make it easy to maintain and its going to be simpler for editors to add articles.

The only missing detail is how to add an image with a caption since it doesn't seem to be a standard in markdown. So, I added a plugin that will use the img alt text as the caption. (You can see it live in the first image on this article `articles/look-mom-no-artist-the-dudes-nfts` (requires reseeding). The image is defined like this `This is a good caption`).

## Checklist

-   [ ] Ready to be merged


resources/js/Pages/Articles/Components/ArticleContent.tsx
```diff
@@ -0,0 +1,14 @@
+import remarkFigureCaption from "@microflash/remark-figure-caption";
+import Markdown from "react-markdown";
+interface Properties {
+    article: App.Data.Articles.ArticleData;
+}
+
+export const ArticleContent = ({ article }: Properties): JSX.Element => {
+    console.log({ article });
```
ItsANameToo, line 8: :watching:

shahin-hq:
It looks beautiful. Nice job. :+1: 
My only comment is about links, I think they should open in a new tab by default. Also we might want to have `nofollow`, `noreferrer` attributes on these links.

ItsANameToo (changes requested):
I updated the articles json to use the markdown alt tag. Issue with that one is that we cannot use links inside of it, which is something they often did for MSQ captions. 

Can you have a look at whether we can make those work in some way to allow links to exist in captions again? I believe that was one of the reasons we went with a custom format on MSQ

alfonsobries:
Replaced the plugin for converting the img alt text to caption in favour of  a custom plugin that converts blockquotes that are next to images to figure with figcaption

So 

```markdown
![](https://i.imgur.com/kw1vnVD.gif)

> the [pixels](http://link.com/page)
```

become

```html
<figure>
  <img src="https://i.imgur.com/kw1vnVD.gif alt="alt text" style="width:100%">
  <figcaption>the <a href="http://link.com/page">pixels</a></figcaption>
</figure>
```

Also links are now opened in a new tab 

cc @ItsANameToo  @shahin-hq

ItsANameToo (approved):

PR #273, opened 2023-10-25
refactor: articles publish date

shahin-hq:
## Summary
https://app.clickup.com/t/86790vqu2

This PR:
- makes `published_at` column `timestamp` instead of `date` because highlighted articles are based on this value, if we publish 2+ articles in a day the order might not be correct because they all will have the same `published_at` value
- makes `sortBy` method consistent with other models
- adds index for the `published_at` as we use this column to sort/filter results
- fixes flaky test

## Checklist

-   [x] Ready to be merged

### DIFF
--- a/tests/App/Http/Controllers/ArticleControllerTest.php
+++ b/tests/App/Http/Controllers/ArticleControllerTest.php
@@ -146,12 +146,12 @@ it('should get featured collections for an article', function () {
 
 it('should keep highlighted articles regardless of the sorting', function () {
     $highlightedArticles = Article::factory(3)->create([
-        'published_at' => now()->addMinutes(10),
+        'published_at' => now()->subMinutes(10),
     ]);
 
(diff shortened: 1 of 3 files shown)


tests/App/Http/Controllers/ArticleControllerTest.php
```diff
@@ -146,12 +146,12 @@ function seedArticles(int $articlesCount = 4, int $collectionsCount = 2): array
 
 it('should keep highlighted articles regardless of the sorting', function () {
     $highlightedArticles = Article::factory(3)->create([
-        'published_at' => now()->addMinutes(10),
```
shahin-hq, line 149: We should extract some minutes here because published means `published_at` < now and  `published_at` is not `null`

crnkovic (approved):

ItsANameToo (approved):

PR #315, opened 2023-10-30
feat: implement gallery drafts hook

shahin-hq:
## Checklist

-   [ ] Ready to be merged


resources/js/Pages/Galleries/hooks/useGalleryDrafts.ts
```diff
@@ -0,0 +1,142 @@
...
+        setIsSaving(false);
+    };
+
+    const getWalletDrafts = async (): Promise<GalleryDraft[]> => {
+        const allDrafts: GalleryDraft[] = await database.getAll();
+
+        return allDrafts.filter((draft) => draft.walletAddress === wallet?.address);
+    };
+
+    const setDraftCover = (image: ArrayBuffer | null, type: string | null): void => {
```
shahin-hq, line 118: We have to use `ArrayBuffer` to store images as Safari cannot store `Blob`s

> If you are storing large, user-generated files such as images or videos, then you may try to store them as File or Blob objects. This will work on some platforms but fail on others. Safari on iOS, in particular, cannot store Blobs in IndexedDB.

> Luckily it is not too difficult to convert a Blob into an ArrayBuffer, and vice versa. Storing ArrayBuffers in IndexedDB is very well supported.

> Remember, however, that a Blob has a MIME type while an ArrayBuffer does not. You will need to store the type alongside the buffer in order to do the conversion correctly.

https://web.dev/articles/indexeddb-best-practices

alfonsobries (changes requested):
code looks good but think we should add some tests, seems complex enough to fail with some minor adjustment

resources/js/Pages/Galleries/hooks/useGalleryDrafts.ts
```diff
@@ -0,0 +1,142 @@
...
+
+const initialGalleryDraft: GalleryDraft = {
+    title: "",
+    cover: null,
+    coverType: null,
+    nfts: [],
+    id: null,
+};
+
+const MAX_DRAFT_LIMIT_PER_WALLET = 6;
```
alfonsobries, line 30: I would move this above the interfaces or below them

alfonsobries:
@shahin-hq maybe we should treat this one and #326 a single task with a single pr, testing this requires #326 anyway

shahin-hq:
@alfonsobries surely we can do that also note that I added unit tests for the hook, so it can be _tested_ by reading unit tests.

alfonsobries:
@shahin-hq think this pr should be against `feat/gallery-drafts`

resources/js/Components/Sidebar/SidebarItem.test.tsx
```diff
@@ -8,6 +8,7 @@ describe("SidebarItem", () => {
             <SidebarItem
                 icon="Cog"
                 title="General"
+                href="/hello"
```
shahin-hq, line 11: had to fix that in order to make CI happy

alfonsobries (approved):

PR #318, opened 2023-10-30
refactor(poc): only index erc721 collections

crnkovic:
## Summary

https://app.clickup.com/t/862kfgzfk

This is a proof of concept. TODO: explain what's happening

## Checklist

-   [ ] Ready to be merged

### DIFF
--- a/app/Http/Client/Mnemonic/MnemonicPendingRequest.php
+++ b/app/Http/Client/Mnemonic/MnemonicPendingRequest.php
@@ -231,6 +231,16 @@ class MnemonicPendingRequest extends PendingRequest
         return intval($data['ownersCount']);
     }
 
+    public function isErc721(Chains $chain, string $contractAddress): bool
+    {
+        $this->chain = MnemonicChain::fromChain($chain);
+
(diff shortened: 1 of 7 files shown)


oXtxNt9U (commented):
I know it's still WIP, but leaving it in case it hasn't been considered:

Do we need a new extra job for this + mnemonic call to determine the type?

Since we use Alchemy to fetch NFTs for a collection, we should be able to determine the collection type from the response already:

image

So alternatively, we could maybe just add a column `token_type` (ERC721, ERC1155, ...) to the collections table instead. And our jobs/queries can filter out collections that are non-ERC721 based on it (meaning we'd keep them in the DB, because they shouldn't really hurt either and might be useful if we change our mind later on to support ERC1155)

Of course this would mean that we at least perform one lookup if it's a collection we haven't seen before, but it shouldn't be much of an issue as long as the `tokenType` is always present.

ItsANameToo:
status on this @crnkovic ?

crnkovic:
@ItsANameToo will take another crack at this and these erc1155 tasks on monday

PR #320, opened 2023-10-30
feat: spread coingecko jobs

alfonsobries:
## Summary

https://app.clickup.com/t/862kf5m7z

<del>only problem is that instead of updating the token details for the top `50` tokes as used to be calculated on the `maxCoingeckoJobsInInterval` method in the `app/Console/Kernel.php` file, it now runs for the top `25` tokens thats because this implementation leaves rooms according to the number of jobs that use coingecko and the max amount of job based on the rate limit (75 every 15 minutes), depending on how we want to be close to the limit this solution may not work and we may need to find a way to "optimize" the spread, im going to revisit tomorrow with a fresh mind to see if there is something I can do, meanwhile in theory task is ready to be tested cc @ItsANameToo </del>

In addition to applying the same logic used with opensea, I had to find a way to optimize how we send jobs. The issue with the implementation we used with opensea is that it considers how many jobs use OpenSea and decides the threshold based on that.

In the case of opensea, we had some leeway because we didn't call it as often, but applying the same implementation for coingecjo meant reducing the number of tokens we can update in a limited time.

So what makes this implementation different is that is also checking which jobs are running at a specific moment. For example, if it's monday after 7 pm, the `VerifySupportedCurrencies` job may be running, which means that only on Monday, we have a higher limit of jobs we can call. On any other day (or earlier), we have more space because we know that this job won't run.

Hope makes sense ping me on slack if something is not clear

## Checklist

-   [ ] Ready to be merged


ItsANameToo (approved):

ItsANameToo:
will further monitor this on demo

oXtxNt9U (approved):

PR #332, opened 2023-11-01
refactor: add dark mode toggle to navbar

crnkovic:
## Summary

https://app.clickup.com/t/862kha6e3

If there's no `theme` key in the localStorage, it will default to whatever your system preferences are. Since the only way to "reverse" back to the system theme after you manually set the theme is to clear localStorage, I suggest we add another option which is "system". So, essentially cycle between light-dark-system when clicking the button.

Dark Mode Context also had a typo in the filename so this PR fixes that.

## Checklist

-   [ ] Ready to be merged


ItsANameToo (changes requested):
image

ItsANameToo (approved):

PR #351, opened 2023-11-03
feat: delete expired drafts

shahin-hq:
## Summary

Closes: https://app.clickup.com/t/862kh23f0

> To avoid stale drafts being kept indefinitely, we should store a timestamp of when they were last edited and discard them from localstorage once they are older than 30 days. This will keep drafts up to date with the changes happening to addresses and avoids needless drafts from piling up.

> A way to handle this is to check upon page navigation to the gallery what the age is of the stored drafts and delete any that are expired before showing the list of drafts

## Checklist

-   [x] Ready to be merged

### DIFF
--- a/resources/js/Pages/Galleries/MyGalleries/Index.tsx
+++ b/resources/js/Pages/Galleries/MyGalleries/Index.tsx
@@ -1,10 +1,11 @@
-import { type ReactNode } from "react";
+import { type ReactNode, useEffect } from "react";
 import { useTranslation } from "react-i18next";
 import { CreateGalleryButton } from "./Components/CreateGalleryButton";
 import Layout from "./Layout";
 import { NftGalleryCard } from "@/Components/Galleries";
 import { Heading } from "@/Components/Heading";
 import { Pagination } from "@/Components/Pagination";
+import { useGalleryDrafts } from "@/Pages/Galleries/hooks/useGalleryDrafts";
 
 const Index = ({
     title,
@@ -21,6 +22,12 @@ const Index = ({
     const { t } = useTranslation();
 
     const userGalleries = galleries.paginated;
+
+    const { deleteExpiredDrafts } = useGalleryDrafts(undefined, true);
+
+    useEffect(() => {
+        void deleteExpiredDrafts();
+    }, []);
 
     return (
         <Layout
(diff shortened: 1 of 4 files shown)


resources/js/databaseConfig.ts
```diff
@@ -11,6 +11,7 @@ export const databaseConfig = {
                 { name: "cover", keypath: "cover", options: { unique: false } },
                 { name: "coverType", keypath: "coverType", options: { unique: false } },
                 { name: "nfts", keypath: "nfts", options: { unique: false } },
+                { name: "createdAt", keypath: "createdAt", options: { unique: false } },
```
crnkovic, line 14: shouldn't this be `updatedAt`?

crnkovic (changes requested):

resources/js/databaseConfig.ts
shahin-hq: Good catch. Should be fine now

ItsANameToo (approved):

resources/js/Pages/Galleries/hooks/useGalleryDrafts.ts
```diff
@@ -131,6 +138,17 @@ export const useGalleryDrafts = (givenDraftId?: number, disabled?: boolean): Gal
         setReachedLimit(false);
     };
 
+    const deleteExpiredDrafts = async (): Promise<void> => {
+        const thirtyDaysAgo = new Date().getTime() - DRAFT_TTL_DAYS * 86400 * 1000;
```
ItsANameToo, line 142: granted this variable name may not be the best since `DRAFT_TTL_DAYS` is configurable 🤔 maybe rename that to `thresholdDaysAgo` or something generic

shahin-hq: Nice, nice. Will fix that :)

alfonsobries (approved):

resources/js/Pages/Galleries/MyGalleries/Index.tsx
```diff
@@ -22,6 +23,12 @@ const Index = ({
 
     const userGalleries = galleries.paginated;
 
+    const { deleteExpiredDrafts } = useGalleryDrafts(undefined, true);
+
+    useEffect(() => {
+        void deleteExpiredDrafts();
+    }, []);
```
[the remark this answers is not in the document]
alfonsobries: I dont think is wrong to delete the expired ones on load which should prevent the rare but possible case where user is seeing the draft page and suddenly data is removed but think we should instead filter the galleries that are loaded from the local storage inside the hook (and eventually remove them when new data is stored), think this approach will show the drafts for a second and then remove it

alfonsobries (commented):
im approving but added a comment that may need discussion

ItsANameToo:
created a card for that @alfonsobries so we can merge this in the meantime: https://app.clickup.com/t/86dqha6ag

PR #442, opened 2023-11-14
refactor: store erc1155 tokens

crnkovic:
## Summary

https://app.clickup.com/t/862kfgzfk + https://app.clickup.com/t/8678v862p

This PR adds a `type` column to the table `collections` table, which indicates the type of a collection (ERC1155 or ERC721). Previously, we had been filtering out non-ERC721 tokens, but since we want to support ERC1155 in the future + show the message to the user that they own ERC1155 tokens, we'll no longer filter them out when returned by the provider and store them. This PR will also show the little "warning" toast if the user owns any ERC1155 tokens.

However, we won't display those ERC1155 collections anywhere and ignore them in jobs and scheduled commands. And if the collection is ERC1155, we won't index its NFTs.

We previously already filtered out non-ERC721, so only the seeded collections can potentially be ERC1155, and as it turns out, there's a lot of them in our dump data (https://github.com/ArdentHQ/dashbrd-data/blob/main/live-dump/collection-nfts/top-eth-collections.json and https://github.com/ArdentHQ/dashbrd-data/blob/main/live-dump/collection-nfts/top-polygon-collections.json, look for `erc_type` column). So this PR will hardcode these existing addresses as they do exist in prod and staging, and set their type to ERC1155.

To determine whether a wallet owns any ERC1155 (to show in the UI), we take their every NFT from the provider and check for ERC1155 type before filtering them out, and update the boolean flag for the user. If it does, the wallet does own it. As we schedule the job to update their NFTs, this will be updated with time. This needs to be done in this way because we don't store ERC1155 NFTs, only the collections -- but the `nfts` table is the link between the wallet and the collection.

### Testing ERC1155 toast

1. Connect the wallet that owns ERC1155 and let it index everything for the wallet
2. Open up Dashbrd and see the toast
3. Click X to hide it
4. Won't show again unless you clear `localStorage`
5. Notice how there are some collections in the database that have ERC1155 type, but they don't have any NFTs indexed

### Testing type

1. `git checkout develop`
2. import the collections dump for prod or staging from Google Drive
3. `git checkout refactor/store-erc1155`
4. `php artisan migrate`
6. Open the database GUI and see how types are set even for existing collections

## Checklist

-   [ ] Ready to be merged


app/Console/Commands/InteractsWithCollections.php
```diff
@@ -34,6 +36,7 @@ public function forEachCollection(Closure $callback, Closure $queryCallback = nu
             ->when($queryCallback !== null, $queryCallback)
             ->select('collections.*')
             ->withoutSpamContracts()
+            ->where('type', TokenType::Erc721)
```
alfonsobries, line 39: I would add a custom scope like scopeErc721() so this can be changed to `->erc721(`

just a suggestion feel free to ignore

app/Data/Wallet/WalletData.php
```diff
@@ -70,6 +74,11 @@ public static function fromModel(Wallet $wallet): self
             ],
             isRefreshingCollections: (bool) $wallet->is_refreshing_collections,
             canRefreshCollections: $wallet->canRefreshCollections(),
+            hasErc1155Nfts: Cache::remember(
+                'wallets:'.$wallet->id.':has-erc1155',
+                now()->addMinutes(30),
+                fn () => $wallet->allNfts()->where('type', TokenType::Erc1155)->exists()
```
alfonsobries, line 80: in case you add the scope this can be updated

resources/js/app.tsx
```diff
@@ -74,7 +75,9 @@ void createInertiaApp({
                             <MetaMaskContextProvider>
                                 <TransactionSliderProvider>
                                     <DarkModeContextProvider>
-                                        <App {...props} />
+                                        <Erc1155ContextProvider>
+                                            <App {...props} />
+                                        </Erc1155ContextProvider>
```
alfonsobries, line 80: just something to think about but wondering if we can get rid of this provider and move the functionality to the `AuthContextProvider` since seems to me something related to the session wallet

WDYT?

alfonsobries (approved):
im approving since seems to work fine, but added some comments that are open to discussion

app/Data/Wallet/WalletData.php
crnkovic: this checks for ERC1155, not ERC721 :trollface: but I'd leave out the scope for this one to be explicit

app/Console/Commands/InteractsWithCollections.php
```diff
@@ -34,6 +35,7 @@ public function forEachCollection(Closure $callback, Closure $queryCallback = nu
             ->when($queryCallback !== null, $queryCallback)
             ->select('collections.*')
             ->withoutSpamContracts()
+            ->onlyErc721()
```
alfonsobries, line 38: not sure about the `only` prefix, in theory I can add a `->orWhere('type', OTHER)` so is not necessary accurate, thoughts?

alfonsobries (approved):

app/Http/Client/Alchemy/AlchemyPendingRequest.php
```diff
@@ -707,6 +708,16 @@ private function getNftV2ApiUrl(): string
         return 'https://'.self::$apiUrlPlaceholder.'.g.alchemy.com/nft/v2/';
     }
 
+    private function getTokenType(string $type): TokenType
```
shahin-hq, line 711: Shouldn't this be in the `app/Enums/TokenType.php`? So you could use it in `app/Http/Client/Moralis/MoralisPendingRequest.php`

database/migrations/2023_11_14_122252_add_type_column_to_collections_table.php
```diff
@@ -0,0 +1,23 @@
...
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\DB;
+use Illuminate\Support\Facades\Schema;
+
+return new class extends Migration
+{
+    public function up(): void
+    {
+        Schema::table('collections', function (Blueprint $table) {
+            $table->string('type')->default(TokenType::Erc721->value);
```
shahin-hq, line 16: Should you add index for this column and it is being used in `where` conditions?

crnkovic: Can't tell for sure, I will conduct experiments on the larger `nfts` table to see where we execute queries to get NFTs, to see if we gain some performance.

app/Http/Client/Alchemy/AlchemyPendingRequest.php
shahin-hq: I agree with that point.

database/migrations/2023_11_14_122252_add_type_column_to_collections_table.php
```diff
@@ -0,0 +1,23 @@
...
+        });
+
+        DB::table('collections')->update([
+            'type' => TokenType::Erc721->value,
```
ItsANameToo, line 20: some seeded collections are ERC-1155, won't this cause them to be incorrectly labeled?

database/migrations/2023_11_14_130544_add_type_column_to_nfts_table.php
```diff
@@ -0,0 +1,26 @@
...
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        Schema::table('nfts', function (Blueprint $table) {
+            $table->string('type')->default(TokenType::Erc721->value);
+        });
+
+        DB::table('nfts')->update([
+            'type' => TokenType::Erc721->value,
```
ItsANameToo, line 23: did this work fine when storing erc1155 tokens? I don't know the exact differences from the API responses when fetching them, only that they differ in nature. If it's stored fine without weird overwrites (E.g. if multiple users own the same kind of ERC1155, how will it handle that as we assume 1 owner at the moment) then we can keep this, otherwise I'd suggest we leave it until we add proper ERC1155 support

database/migrations/2023_11_14_122252_add_type_column_to_collections_table.php
ItsANameToo: any update on this?

database/migrations/2023_11_14_130544_add_type_column_to_nfts_table.php
ItsANameToo: an example https://rarible.com/token/0x488130618bf46cd904fdfe1e2f94d377d73f5eb9:1 , only 1 token but there are 311 owners (22 remaining unowned). How would that be handled if we have address A that owns it and connects to dashbrd, but address B that also owns that NFT which connects at a later time?

ItsANameToo (changes requested):

database/migrations/2023_11_14_122252_add_type_column_to_collections_table.php
crnkovic: i'll look which of the seeded are erc1155 and hardcode this for them

database/migrations/2023_11_14_130544_add_type_column_to_nfts_table.php
crnkovic: interesting, i might have overlooked the fact that erc1155 can have multiple owners, and we'd need to map nft to multiple wallets

oXtxNt9U: @crnkovic For multiple owners, see this task https://app.clickup.com/t/86dqnjq6h but that's for a follow up

ItsANameToo:
some conflicts due to the alchemy changes @crnkovic

app/Models/User.php
```diff
@@ -113,7 +114,7 @@ public function nfts(): HasManyThrough
      */
     public function collections(): Builder
     {
-        return Collection::whereIn('collections.id', function ($query) {
+        return Collection::where('collections.type', TokenType::Erc721)->whereIn('collections.id', function ($query) {
```
alfonsobries, line 117:
```suggestion
        return Collection::erc721()->whereIn('collections.id', function ($query) {
```

oXtxNt9U (approved):
@crnkovic  see alfy's feedback

otherwise LGTM

ItsANameToo (approved):

PR #471, opened 2023-11-17
refactor: migrate `getNftsForCollection` to Alchemy v3

shahin-hq:
## Summary
Depends on #468 

Closes: https://app.clickup.com/t/86dqjv2nt

## Checklist

-   [ ] Ready to be merged


shahin-hq:
@goga-m Please read the task description :)

ItsANameToo:
remove v2 as discussed

crnkovic (approved):

app/Http/Client/Alchemy/AlchemyPendingRequest.php
```diff
@@ -360,6 +360,97 @@ public function getContractMetadataBatch(array $contactAddresses, Network $netwo
         });
     }
 
+    /**
+     * @param  array<mixed>  $nft
+     */
+    public function parseNftV3(array $nft, int $networkId, bool $convertTokenNumber = true): Web3NftData
```
ItsANameToo, line 366:
```suggestion
    public function parseNft(array $nft, int $networkId, bool $convertTokenNumber = true): Web3NftData
```

ItsANameToo: applies to all `V3` names

tests/fixtures/alchemy/get_nfts_for_contract.json
```diff
@@ -0,0 +1,293 @@
+{
+    "nfts": [
```
ItsANameToo, line 2: general remark here as well, do we have any error states that can occur in the API responses? would be good to have tests for those as well

shahin-hq: You mean error states with non-200 status responses or responses with `raw.error` is not null? If I am not mistaken non-200 responses already handled in the `send` [method](https://github.com/ArdentHQ/dashbrd/blob/70ed02390f45f6d0056f10092c90bd31a04a02fd/app/Http/Client/Alchemy/AlchemyPendingRequest.php#L96), the responses with `raw.error` should already have it. Otherwise we would have coverage issue.

ItsANameToo: yeah actual errors with `raw.error`. Main question is: are V3 error responses the same as V2 or not?

ItsANameToo: if is the same, then good to go, just want confirmation of that

shahin-hq: image
According to spec the `error` key just moved into the `raw`, so I think we should be fine.

PR #484, opened 2023-11-21
refactor: remove references to featured collections

crnkovic:
## Summary

Now that we'll have an actual concept of "featured" collections, I thought it might be beneficial for us to remove references to "old" featured collections, which are not "featured" at all - but rather just collections assigned to an article.

We had this `withFeaturedCollections` scope on the article model, and that will 100% cause issues in the future once we add featured collections, since this doesn't load just featured collections, but just eager loads all collections for an article.

## Checklist

-   [ ] Ready to be merged

### DIFF
--- a/app/Models/Article.php
+++ b/app/Models/Article.php
@@ -36,7 +36,7 @@ class Article extends Model implements HasMedia, Viewable
 
     public function resolveRouteBinding($value, $field = null)
     {
-        return Article::query()->withFeaturedCollections()->where('articles.slug', $value)->first();
+        return Article::query()->with('collections')->where('articles.slug', $value)->first();
     }
 
     public function registerMediaCollections(): void
@@ -136,21 +136,6 @@ class Article extends Model implements HasMedia, Viewable
         $nullsPosition = Str::lower($direction) === 'asc' ? 'NULLS FIRST' : 'NULLS LAST';
 
         return $query->orderByRaw("articles.published_at {$direction} {$nullsPosition}");
-    }
-
-    /**
-     * @param  Builder<self>  $query
-     * @return Builder<self>
-     */
-    public function scopeWithFeaturedCollections(Builder $query): Builder
-    {
-        return $query->with(['collections' => function ($query) {
-            $query->select([
-                'collections.name',
-                'collections.slug',
-                'collections.extra_attributes->image as image',
-            ]);
-        }]);
     }
 
     public function metaDescription(): string
--- a/app/Data/Articles/ArticleData.php
+++ b/app/Data/Articles/ArticleData.php
@@ -19,7 +19,7 @@ use Spatie\TypeScriptTransformer\Attributes\TypeScript;
 class ArticleData extends Data
 {
     /**
-     * @param  DataCollection<int, FeaturedCollectionData>  $featuredCollections
+     * @param  DataCollection<int, FeaturedCollectionData>  $collections
      */
     public function __construct(
         public int $id,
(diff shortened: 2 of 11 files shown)


app/Models/Article.php
```diff
@@ -138,21 +138,6 @@ public function scopeSortByPublishedDate(Builder $query, string $direction = 'de
...
-     */
-    public function scopeWithFeaturedCollections(Builder $query): Builder
-    {
-        return $query->with(['collections' => function ($query) {
-            $query->select([
-                'collections.name',
-                'collections.slug',
-                'collections.extra_attributes->image as image',
-            ]);
-        }]);
```
alfonsobries, line 153: this scope used to select only those columns, what if rename the scope to `withRelatedCollections` or something like that so we dont lose the select

alfonsobries (changes requested):

app/Models/Article.php
crnkovic: I can add that, i just didn't see any value in selecting only few columns - in fact i see something that's bothering me, and that's the fact that we're selecting `extra_attributes->image` as `image`, and then use `$model->image` to reference that "imaginary" field

I don't think these imaginary attributes are a good approach, especially with the lack of type safety with Laravel

shahin-hq: > i just didn't see any value in selecting only few columns
Why? You would like to select all columns when you don't need?

crnkovic: I don't see why not, what is the issue with selecting all columns in this case?

shahin-hq: It is not a good approach to select everything when you don't need:

image

crnkovic: Sure, some random page on the internet says that. In practice and in this concrete example, why is bad to select all columns for a table with 1000 records and 25 columns? Don't you think that having some virtual, imaginary property on a model is objectively worse than selecting all columns from such a tiny table? Let's not make the code less readable and less maintainable for the sake of literal micro-optimizations.

I can give at least 3 examples related to database optimizations that are more important than this change, yet I didn't see you commenting on that. Let's not nitpick.

shahin-hq: > Sure, some random page on the internet says that. In practice and in this concrete example, why is bad to select all columns for a table with 1000 records and 25 columns?

I can give you more credible source if you want, I've read it many times:
High Performance MySQL: Optimization, Backups, and Replication Third Edition: [link](https://gateway.ipfs.io/ipfs/bafykbzacebhf4eml4bpzymd7zwoothcyr6j54mwvez7k335abio32cl7ewklm?filename=Baron%20Schwartz%2C%20Peter%20Zaitsev%2C%20Vadim%20Tkachenko%20-%20High%20performance%20MySQL_%20optimization%2C%20backups%2C%20and%20replication-O%E2%80%99Reilly%20Media%20%282012%29.pdf)
Chapter 6: Query Performance Optimization page 203
image


> Don't you think that having some virtual, imaginary property on a model is objectively worse than selecting all columns from such a tiny table?

That is a different problem than above, why would follow good approaches in one direction but not in another, if good practices say not to use `virtual` columns, using a proper name would solve this.

I am not nitpick-ing, if we are following good practices, we should do it when possible. No hard feelings.

crnkovic: I was more talking about the fact that just because it's said somewhere doesn't mean you have to blindly follow it for the sake of just "doing best practices" for truly negligible performance improvement of 10-15 milliseconds when it can also be handled with caching -- especially in the early MVP stage where tables are modified a bunch and we have to run back every 5th PR to change which columns are selected where.

Also, there's no need to explain that selecting all columns is performing worse than selecting a few columns -- judging from the dozen of my PRs that heavily optimized SQL queries in this project, I'm very sure I don't need to be educated on how to optimize the queries that matter.

And yes, this was nitpicking.

You're talking about two separate problems, which are indeed there. My "Don't you think" was about the fact that you didn't nitpick in such a way when this "aliasing and then assigning to a magic class property" was introduced. Please, let's focus on more important things.

crnkovic: @alfonsobries will push your suggestions

alfonsobries (approved):

PR #493, opened 2023-11-23
refactor(proof of concept): store when NFTs are burned

crnkovic:
## Checklist

-   [ ] Ready to be merged

### DIFF
--- a/app/Jobs/FetchCollectionActivity.php
+++ b/app/Jobs/FetchCollectionActivity.php
@@ -118,7 +118,7 @@ class FetchCollectionActivity implements ShouldQueue
         $burnActivities = $activities->filter(fn ($activity) => $activity->type === NftTransferType::Burn);
 
         if ($burnActivities->isNotEmpty()) {
-            SyncBurnedNfts::dispatch($burnActivities);
+            SyncBurnedNfts::dispatch($this->collection, $burnActivities);
         }
 
         DB::transaction(function () use ($formattedActivities, $activities) {
--- a/app/Jobs/FetchCollectionActivity.php
+++ b/app/Jobs/FetchCollectionActivity.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace App\Jobs;
 
 use App\Data\Web3\CollectionActivity;
+use App\Enums\NftTransferType;
 use App\Jobs\Traits\RecoversFromProviderErrors;
 use App\Models\Collection;
 use App\Models\NftActivity;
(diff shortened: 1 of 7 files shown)


alfonsobries (approved):
In terms of code, it looks good, and I think it's a relevant detail to have. My only consideration is that we should only add this PR if it's really going to be used for any of the functions you mention like updating NFT metadata, (at least have some tasks in ClickUp to implement it) otherwise I would wait to implement this.

In short: in my opinion, I believe it's an improvement and starting by updating the NFT metadata only for non-burned NFTs seems like a good task to add

crnkovic:
@alfonsobries applied one use case which is excluding burned NFTs from refreshing NFT metadata. this can potentially run for a lot of NFTs, so this approach might speed up the job runtime. looking into what else i can improve with this burned NFTs concept, but wanna keep the PR light, so will create a card when I discover other use-cases

oXtxNt9U (changes requested):
How do we apply this retroactively? I think we need some kind of migration job that loops over any existing activities to sync the burn column.

Might be good to check with real data how many affected NFTs we'd even have (just to have a general idea).

Additionally, a burn generally means that there's no owner anymore. So to be consistent we probably need to set `wallet_id` to null here or else we might end up with burned NFTs that still have a wallet attached (tbf it should fix itself eventually):

https://github.com/ArdentHQ/dashbrd/pull/493/files#diff-4b4b4cff5fbc12e1e63c802626218821b6e915b034588fe018272513129a4ffbR58

crnkovic:
> How do we apply this retroactively? I think we need some kind of migration job that loops over any existing activities to sync the burn column.

True, would need a job to go through collections, fetch the activity, and store `LABEL_BURN` events. Will probably be a lot faster, as we only care about this type. Will test and add it.

crnkovic:
Testing on the collections dump:

- 248/554 collections from demo have burn events, in total of around 116k events, took around 30mins to update all
- 181/417 collections from prod have burn events, in total of around ~500k events, took around 1hr to update all

crnkovic:
> Though not sure if it makes a difference in practice, maybe it's fine to just not bother since it should correct itself anyway after a FetchWalletNfts job.

I agree, this will clear itself. I feel this will require much more testing as we remove `wallet_id` and I'm kinda scared to push that in this PR, since this PR is already pretty beefy.

oXtxNt9U (approved):

ItsANameToo (approved):

PR #490, opened 2023-11-22
feat: add top/floor price tabs for filtering

crnkovic:
## Summary

https://app.clickup.com/t/86dqk9j3e

Built from #487 

## Checklist

-   [ ] I checked my UI changes against the design and there are no notable differences
-   [ ] I checked my UI changes for any responsiveness issues
-   [ ] I checked my (code) changes for obvious issues, debug statements and commented code
-   [ ] I opened a corresponding card on Clickup for any remaining TODOs in my code
-   [ ] I added a short description on how to test this PR _(if necessary)_
-   [ ] I added a storybook entry for the component that was added _(if necessary)_
-   [ ] Documentation _(if necessary)_
-   [ ] Tests _(if necessary)_
-   [ ] Ready to be merged

PR #497, opened 2023-11-23
Feat/mobile filters

alfonsobries:
## Summary

Closes https://app.clickup.com/t/86dqm6qep

Depends on https://github.com/ArdentHQ/dashbrd/pull/490

## Checklist

-   [ ] Ready to be merged

### DIFF
--- a/resources/js/Pages/Collections/Index.tsx
+++ b/resources/js/Pages/Collections/Index.tsx
@@ -1,6 +1,7 @@
 import { type PageProps } from "@inertiajs/core";
 import { Head, usePage } from "@inertiajs/react";
 import { useTranslation } from "react-i18next";
+import { PopularCollectionsFilterPopover } from "./Components/PopularCollectionsFilterPopover";
 import { PopularCollectionsSorting } from "./Components/PopularCollectionsSorting";
 import { PopularCollectionsTable } from "@/Components/Collections/PopularCollectionsTable";
 import { Heading } from "@/Components/Heading";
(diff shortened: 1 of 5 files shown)


ItsANameToo:
conflicts after merging #490

crnkovic (changes requested):
There are some conflicts; also can you resolve the width issue on tabs on 640px and 768px:

Screenshot 2023-11-27 at 11 30 39

resources/js/Pages/Collections/Index.tsx
```diff
@@ -31,22 +34,45 @@ const CollectionsIndex = ({
...
+        for (const key of keys) {
+            const value = currentFilters[key];
+
+            if (value === null) continue;
+
+            query[key] = value;
+        }
+
+        router.get(route("collections"), query, {
+            only: ["collections", "activeSort", "filters"],
```
shahin-hq, line 57: Looks good to me. Only thing is using `only` here will not load data lazily as backend implementation doesn't have `closures` in place so we could omit this property. 

Beside that we can use `router.reload()` over `router.get()` as it will use the current URL and make this call a little bit simpler.

Ref here: https://inertiajs.com/partial-reloads

alfonsobries:
@shahin-hq @crnkovic updated

crnkovic (approved):
There was a little bug where active sorting was not properly highlighted in the tab if you didn't activate any sorting (sort by "Top"). I found that the bug was related to you overriding [`sort` property](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort) on the empty array (when no filters are there), so I pushed a little fix to cast "filters" to object from the controller, so that it's always a JavaScript object in the frontend, and not an array when there are no filters applied.

In essence, the bug was that this property was always a function on an empty array, as per MDN, but we compared it to `undefined` when no filters are selected.

PR #548, opened 2023-12-09
feat: add volume metrics

alfonsobries:
## Summary

Closes https://app.clickup.com/t/86dqrnjn8

Since we already fetch the volume for the collection, I modified the mnemonic request to be compatible with [HTTP concurrent requests](https://laravel.com/docs/10.x/http-client#concurrent-requests) in order to fetch the rest of the periods we need concurrently.

With this change, we should be able to fetch all values with a single call to the request <del>(I still need to analyze the rate-limit implications)</del>.

<del>P.S.: The PR is pretty much working, but I wasn't able to mark it as finished because I'm still adapting the tests for the pool call (Mnemonic::fake doesn't seem to work).</del>

<del>Meanwhile, it would be great to hear your opinions on this approach, @crnkovic, @ItsANameToo, @shahin-hq . Also, if you want, feel free to take over this PR in case there are no more tasks to tackle. Otherwise, I will continue once I'm at my desk again</del>

### Update: Rate limit Implications:

The volume is updated every time we:

a. Call the `SyncCollection` job. This occurs when the collection is viewed for the first time after 1 day (see `Collection@recentlyViewed` method), so it's a maximum of 1 time per day (per collection).
b. When the `RefreshWalletCollections` job is called (manually by the user), a maximum of once every 15 minutes (per collection).

In both jobs, `SyncCollection` and `RefreshWalletCollections`, we also update some other data with other jobs (floor price, traits, owners, and currently a single volume). IMO updating the rest of the volume metrics there make sense which means we need to make 2 extra api calls on whatever approach we choose, also since both jobs are not called every time and more like after user interactions I dont think that the adding two extra calls is going to represent a real impact to our rate limits for now.

Idea: What comes to my mind (either with this adjustment or without it) is to update the FetchCollectionVolume job to ensure it only updates the volume once a day in case it is called multiple times. We can simply add a job rate limiter for that. What do you think? If so, we can manage that as a separate task.

## Checklist

-   [ ] Ready to be merged


crnkovic:
Very interesting approach. My initial thoughts were that this was going to be the same as logging floor price changes, as it's essentially the same: logging some change once a day, then calculating changes/percentages over 24hr/1w/1m.

oXtxNt9U:
> @crnkovic Very interesting approach. My initial thoughts were that this was going to be the same as logging floor price changes (log every time we update volume), as it's essentially the same: logging some change once a day, then calculating changes/percentages over 24hr/1w/1m.

This only works if we already have daily stats, but initially when we see a collection for the first time there'd be no data to calculate the 1 week/1 month numbers.

However, if we already have the daily volumes for the past 7 / 30 days, then yes we could do that instead. On top of that it would be interesting to know how much our calculated volumes would differ from the mnemonic reported stats (should be very close) i.e. calculating 1 week / 30 day from daily stats vs. mnemonic response

So yeah I think we can potentially stop doing the 2 extra requests for collections if they have been indexed for more than 7 / 30 days.

> @alfonsobries  Idea: What comes to my mind (either with this adjustment or without it) is to update the FetchCollectionVolume job to ensure it only updates the volume once a day in case it is called multiple times. We can simply add a job rate limiter for that. What do you think? If so, we can manage that as a separate task.

Not sure about that :thinking: Having the volume potentially lag behind for an entire day sounds not ideal.


Pinging for additional input @ItsANameToo

crnkovic:
P.S. I took a different approach in this PR (https://github.com/ArdentHQ/dashbrd/pull/576) so we can discuss both

ItsANameToo:
@alfonsobries please update the PR description with the amount of requests needed + a short overview of monthly compute costs and how that translates to x amount of collections (e.g. 500, 5000)

crnkovic:
For the amount of API requests, it would just be 3x the average Mnemonic requests. You can see the number of requests per month by just signing into the Mnemonic console (https://app.mnemonichq.com/plans). However, with this PR, it may result in stale data if volume for some collections hasn't been updated in a while.

crnkovic:
#576 was merged, so closing this

PR #554, opened 2023-12-11
feat: floor price change percentage

alfonsobries:
## Summary

Closes https://app.clickup.com/t/86dqrnkyz

- adds price history table and store new entries every time a new floor price is fetched
- Calculates the floor price change by using the average price from today vs the average day from yesterday
- updates the frontend to reflect those percentages and shows `N/A` if we dont have the data
- im not considering any kind of pruning for now, we can handle that if table starts to grow up too much

## Checklist

-   [ ] Ready to be merged


database/migrations/2023_12_11_202944_create_floor_price_history_table.php
```diff
@@ -0,0 +1,23 @@
...
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+return new class extends Migration
+{
+    public function up(): void
+    {
+        Schema::create('floor_price_history', function (Blueprint $table) {
+            $table->id();
+            $table->foreignIdFor(Collection::class)->constrained()->cascadeOnDelete();
```
oXtxNt9U, line 17: We should make sure this column has an index (and also `retrieved_at`)

database/migrations/2023_12_11_202944_create_floor_price_history_table.php
```diff
@@ -0,0 +1,23 @@
...
+use App\Models\Token;
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+return new class extends Migration
+{
+    public function up(): void
+    {
+        Schema::create('floor_price_history', function (Blueprint $table) {
```
oXtxNt9U, line 15: Does it make sense for `floor_price` and `token_id` to be nullable? Couldn't we instead just not store anything if we don't have either of them?

oXtxNt9U (changes requested):

database/migrations/2023_12_11_202944_create_floor_price_history_table.php
alfonsobries: @oXtxNt9U yes, good catch

crnkovic (commented):
I like this approach. One question: does the `retrieved_at` column need to be `nullable`? Because essentially, `retrieved_at` is current time, right? I know we had some logic for null values in the job class, but can we ditch that? Can `retrieved_at` be anything else other than `NOW()` when storing? I don't understand why we remove time if token is missing (nothing with your PR, just thinking out loud as we do that prior to this PR as well).

Also, if `retrieved_at` is `null`, have you tested that it calculates values properly? Since you compare with `CURRENT_DATE`, etc.

alfonsobries:
@crnkovic retrieved_at shouldnt be null, just fixed it

crnkovic (approved):

oXtxNt9U (approved):

crnkovic:
Added the task for pruning old changes: https://app.clickup.com/t/86dqum8fj

PR #565, opened 2023-12-14
refactor(proof of concept): aggregate "collection of the month" winners once a month

crnkovic:
## Checklist

-   [ ] Ready to be merged

### DIFF
--- a/app/Jobs/AggregateCollectionWinners.php
+++ b/app/Jobs/AggregateCollectionWinners.php
@@ -1,0 +1,60 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Jobs;
+
+use App\Models\Collection;
+use App\Models\CollectionWinner;
+use Illuminate\Bus\Queueable;
+use Illuminate\Contracts\Queue\ShouldBeUnique;
+use Illuminate\Contracts\Queue\ShouldQueue;
+use Illuminate\Foundation\Bus\Dispatchable;
+use Illuminate\Queue\InteractsWithQueue;
+use Illuminate\Queue\SerializesModels;
+use Illuminate\Support\Collection as LaravelCollection;
+
+class AggregateCollectionWinners implements ShouldBeUnique, ShouldQueue
+{
+    use Dispatchable, InteractsWithQueue, Queueable, SerializesModels;
+
+    /**
+     * Execute the job.
+     */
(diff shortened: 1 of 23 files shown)


app/Jobs/AggregateCollectionWinners.php
```diff
@@ -0,0 +1,60 @@
...
+    /**
+     * Execute the job.
+     */
+    public function handle(): void
+    {
+        $previousMonth = now()->subMonth();
+
+        $winners = $this->winners();
+
+        CollectionWinner::insert($winners->map(fn ($collection) => [
```
oXtxNt9U, line 30: If this job runs multiple times within the same month (accidentally executed manually, a retry or for whatever other reason) we can end up with more than 3 winners since this doesn't use any unique index or similar. The same collection could even appear as a winner more than once in the same (month, year).

Maybe instead we could only store a single row for a (month, year) with a unique index here
plus 3 extra columns to store the 1st, 2nd and 3rd place collection ids.

Then on insert, we would on conflict either
- do nothing
- upsert and overwrite the collections we previously set

whichever we chose, this way it's not possible to end up with duplicates or more than 3 winners in a month.

app/Jobs/AggregateCollectionWinners.php
```diff
@@ -0,0 +1,73 @@
...
+        });
+    }
+
+    /**
+     * @return LaravelCollection<int, Collection>
+     */
+    private function winners(): LaravelCollection
+    {
+        return Collection::query()
+                        ->withTrashed()
```
shahin-hq, line 59: I wouldn't include trashed collections as we won't be able to show them in the collections page, wdyt?

resources/js/Pages/Collections/Components/WinnerCollections/WinnerCollections.blocks.tsx
```diff
@@ -315,14 +315,27 @@ export const WinnerCollectionsTable = ({
...
+                            "March",
+                            "April",
+                            "May",
+                            "June",
+                            "July",
+                            "August",
+                            "September",
+                            "October",
+                            "November",
+                            "December",
```
shahin-hq, line 330: I'd probably move these to the translation file and use index to get it

oXtxNt9U (approved):

PR #657, opened 2024-02-08
refactor: remove double call to fetch opensea collection slugs

crnkovic:
## Summary

https://app.clickup.com/t/86drg8y72

## Checklist

-   [ ] Ready to be merged


app/Jobs/FetchCollectionOpenseaSlug.php
```diff
@@ -37,10 +36,6 @@ public function __construct(
      */
     public function handle(): void
     {
-        Log::info('FetchCollectionOpenseaSlug Job: Processing', [
-            'collection' => $this->collection->address,
-        ]);
```
ItsANameToo, line 42: these were added to get some insights into the job (when it runs, what it retrieves). What was the reason for removing them?

crnkovic: I'd much rather only leave the "worst case scenario" (no slug has been found). We have logrotate which will remove the logs after some time, and if we log every path that every job can take, then the logs will be rotated even faster, making logging useless.

I've updated the PR to log if OpenSea slug is not found (much more useful to know).

PR #687, opened 2024-03-08
refactor: only show erc721 collections in various tables

crnkovic:
## Summary

https://app.clickup.com/t/86dru9vy6

Since we throw 404 when you try to open an ERC1155 collection, it makes the most sense to just exclude those collections from being listed. On top of that, currently we only store ERC1155 collections to determine whether the user 

This change will apply to:
- on `/collections/popular` page
- on `/my-collections` page
- on the `/collections` page under:
    - Popular Collections table
    - "Vote for Top Collection of the Month" table
    - dialog when searching collections to nominate

To test, find some random collection that's shown in these tables, open the database, and change the `type` column for that collection to `erc1155`. After refreshing the page, that collection won't be shown anymore.

## Checklist

-   [ ] Ready to be merged

### DIFF
--- a/app/Http/Controllers/Api/NominatableCollectionController.php
+++ b/app/Http/Controllers/Api/NominatableCollectionController.php
@@ -28,6 +28,7 @@ class NominatableCollectionController extends Controller
                                 ->searchByName($request->get('query'))
                                 ->limit(15)
                                 ->votable()
+                                ->erc721()
                                 ->withCount(['votes' => fn ($q) => $q->inCurrentMonth()])
                                 ->orderBy('votes_count', 'desc')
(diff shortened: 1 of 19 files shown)


app/Http/Controllers/Api/NominatableCollectionController.php
```diff
@@ -28,6 +28,7 @@ public function index(Request $request): JsonResponse
                                 ->searchByName($request->get('query'))
                                 ->limit(15)
                                 ->votable()
+                                ->erc721()
```
ItsANameToo, line 31: could we set a default scope to only use erc721? and have on explicitly to include erc1155? that might avoid these issues going forward instead of having to explicitly ignore the ones we don't want to show 🤔

